PR #3, opened 2015-04-08
Allow prediction of probabilities and selection of scoring method.

pfsq:
Hi Sebastian,

I've started to use your extensions to assess my models before making submissions to Kaggle and found it quite useful so far. But in order to match Kaggle's evaluation methods for different competitions I added some functionality to the learning_curves function.

Basically, I added `predict_proba` to predict probabilities and `scoring` to select the metrics.

Regards.

### DIFF
--- a/mlxtend/evaluate/learning_curves.py
+++ b/mlxtend/evaluate/learning_curves.py
@@ -71,7 +71,7 @@ def plot_learning_curves(X_train, y_train, X_test, y_test, clf, kind='training_s
     if kind == 'training_size':
         rng = [int(i) for i in np.linspace(0, X_train.shape[0], 11)][1:][::-1]
         for r in rng:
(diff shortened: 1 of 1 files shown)


rasbt:
Hi, Pablo,

thanks a lot for adding those features! I made a few changes though. I didn't want to make this function 100% dependent on scikit-learn thus the sklearn.metrics is not imported by default, only if a different metric than "missclassification error" is chosen. 
Somehow I couldn't get the `predict_proba` to work and removed it for now. Unfortunately, I don't have much time for tinkering right now due to my work and book writing in my free time. But it would be nice if you could add the `predict_proba` with an example in `mlxtend/tests/tests_evaluate/test_learning_curves.py` and/or IPython notebook at `docs/examples/evaluate_plot_learning_curves.ipynb`

Anyway, good luck for the competition!

Cheers,
Sebastian

PR #89, opened 2016-08-11
change max_score - line 243

wahutch:
Change max_score in line 243 to accommodate smaller negative values. For example, when using 'mean_squared_error', the call to get_scorer results in a make_scorer function with 'greater_is_better = False', such that values less than -1 can occur (the sign is flipped on the result of mse). Changed to float('-inf'), or could be -10**9, etc.

### DIFF
--- a/mlxtend/feature_selection/sequential_feature_selector.py
+++ b/mlxtend/feature_selection/sequential_feature_selector.py
@@ -240,7 +240,7 @@ class SequentialFeatureSelector(BaseEstimator, MetaEstimatorMixin):
                 sys.stderr.flush()
 
         if select_in_range:
-            max_score = -1
+            max_score = float('-inf')
             for k in self.subsets_:
                 if self.subsets_[k]['avg_score'] > max_score:
                     max_score = self.subsets_[k]['avg_score']
--- a/mlxtend/feature_selection/tests/test_sequential_feature_selector.py
+++ b/mlxtend/feature_selection/tests/test_sequential_feature_selector.py
@@ -372,6 +372,7 @@ def test_regression():
                 skip_if_stuck=True,
                 print_progress=False)
     sfs_r = sfs_r.fit(X, y)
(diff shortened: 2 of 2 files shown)


rasbt:
That's awesome, thanks! Would you mind adding a unit test for such a scenario (negative scores) before I merge? That would be doubly-great!

wahutch:
I'd be happy to do that! I'm very new to git, unit testing, etc. I'll give it a go:)

rasbt:
Thanks! The unit tests for this feature selector are in the `mlxtend/feature_selection/tests/test_sequential_feature_selector.py` file, and if you run nosetests, I wouldn't run it on the complete mlxtend directory (I am not sure if you have tensorflow installed and these tests may fail in this case then). However, if you do want to run nose on the complete mlxtend package, I recommend running it with the `nosetests --nologcapture` flag because the tensorflow code produces a lot of distracting log-output.

wahutch:
OK, I added a new function for testing a range for k_features in combination with a regression loss function such as mse. It may be redundant in that you could just change the existing test_regression function to have a tuple input for k_features. Let me know if you had something else in mind. 

Very impressive (and useful) work on this package!

rasbt:
Thank a lot for your help! Looks good to me. I think it's good to leave both tests in there to test that both scenarios (range, no range) work properly. A small tweak for clarity: Could you add to little assertions. E.g., for the first one

    assert len(sfs_r.k_feature_idx_) == 13

and for the second one (the range):

    assert len(sfs_r.k_feature_idx_) == {some number that is prob. < 13}

So that people who look at the unit tests don't get confused why the scores are different.

Ah, and a very minor thing: could you please insert a space in 

 
    ...
    k_features=(1,13),
    ...

so that
 
    ...
    k_features=(1, 13),
    ...

wahutch:
Yes, definitely. Great idea to add those assertions. The distinction is much clearer that way.

rasbt:
That's great, thanks for the contribution!

PR #107, opened 2016-09-27
Stoppable Sequential Feature Selection

wdm0006:
Large SFS jobs can be pretty slow at times, so it’s useful to track not only the number of features that have been selected (the current print_progress), but also the avg. score at that step, and allow the user to control+c to raise a keyboard interrupt, and then catch and handle that safely to get back the subset of selected features.  

This PR accomplishes these by adding an optional verbose parameter (similar to other libraries), which allows the user to get not only the existing progress, but also to raise the verbosity to get detailed progress updates like:

   [2016-09-27 15:45:52] Features: 1/50 -- avg. score: 0.566246124946
   [2016-09-27 15:46:29] Features: 2/50 -- avg. score: 0.592720840564
   [2016-09-27 15:47:06] Features: 3/50 -- avg. score: 0.606385302233
   [2016-09-27 15:47:46] Features: 4/50 -- avg. score: 0.625515735334
   [2016-09-27 15:48:26] Features: 5/50 -- avg. score: 0.633775958769

And when the user is happy with the score, or unhappy with the runtime, can keyboard interrupt and safely return the data backing the 5 features that have been selected so far.

No core functionality of the feature selection is changed.


rasbt:
Thanks for the pull request, this looks like a super useful feature! In this case, I think it may be a good idea to deprecate `print_progress` and control the printed output by the `verbose` parameter only (for simplicity). I can do this separately later; alternatively, feel free to add the deprecation to this pull request. Also, is there a chance adding a unittest, e.g., using `signal.SIGINT` like shown here: http://stackoverflow.com/questions/32023719/how-to-simulate-a-terminal-ctrl-c-event-from-a-unittest

wdm0006:
Sounds good, I'll go ahead and deprecate print_progress and add a unit test.

rasbt:
Great, thanks! Also, would be nice if you could add a note to the [Changelog](https://github.com/rasbt/mlxtend/blob/master/docs/sources/CHANGELOG.md); maybe under "new features"

wdm0006:
So I've updated the changelog and deprecated print_progress locally, and have made some progress on the test, but it's not really working great, and if I solve some reliability issues in the test, it would still depend on a `time.sleep(N)` that is set correctly to kill it not too late / not too early.  

Having gone through it now, I'm not sure the unit test will solve any more headaches than it will make. I can keep working on it if you'd like, but otherwise can push the other changes now.

rasbt:
> it would still depend on a time.sleep(N) that is set correctly to kill it not too late / not too early.

Yeah, that would be a bit tricky since we don't know the performance of the Appveyor and Travis CI servers, and it could cause some headaches, indeed. 

Hm, but what do you think about an additional parameter `self.interrupted_=False` that is set to `True` if the SFS is interrupted? Other than unit testing, it may be useful as an identifier in general. Then say the unit test runs 10 sec and the interruption occurs at 0.1 sec or so just to check that this feature works in general, which could be useful for further modifications of the code in future,.

wdm0006:
Yeah, that's a good idea, I'll work on that.

wdm0006:
Ok, so it's not super pretty, but a reliable way to test this behavior without sending SIGINTs around and running things in subprocesses is to just add a protected instance variable to the class: `self._TESTING_INTERUPT_MODE=False`, and if that's true, explicitly raise a keyboard interrupt after the first iteration, triggering the safe handling. 

That's super easy to test, and will work on any system. 

I also added in the `self.interrupted_` parameter that you suggested, and tested that as well.

Let me know if you think that's alright.

rasbt:
Independent from your contribution, I just noticed that there are 3 unit tests that are failing due to the release of sklearn 0.18 today. It's not big deal to fix them though, basically, the problem was in one of the unit testing suits since the `NotFitterError` was moved to `sklearn.exceptions`. However, since I am at it, I will get rid of scikit's deprecation warning in one swoop and make the code compatible to both 0.17 and 0.18. Will let you know when it's done (probably tonight or tomorrow) since you probably need to rebase afterwards. Sorry about the inconvenience!

rasbt:
Just merged #110 and it should all be compatible to both scikit-learn 0.17 and 0.18 now; hope it doesn't cause too much trouble for this PR

wdm0006:
Pulled in the updates from master, doesn't seem to have effected much for this.

rasbt:
Thanks again for the tweaks! However, it seems that the integration tests fail due to a missing `print_progress` parameter in one of the tests (line 337). Could you update the unit tests? 

Instead of removing `print_progress`, I think it would be better to deprecate it. I think sth like the following should work:


```python
import warnings


def __init__(self, 
             estimator, k_features='best',
             forward=False, print_progress=False, verbose=1, 
             floating=False, scoring=None,
             cv=5, skip_if_stuck=True, n_jobs=1,
             pre_dispatch='2*n_jobs',
             clone_estimator=True):
    
    # ...
    if print_progess:
        warnings.warn("The print_progress parameter "
                      "has been deprecated in "
                      "0.4.3 and will be removed in 0.5.0. "
                      "Please use the verbose parameter instead.",
                      DeprecationWarning)
        verbose = 1

    # ...
    self.verbose = verbose
    # ...    
```

wdm0006:
Sounds good to me.

rasbt:
That's great, thanks! Was just playing around with it ... sorry, one more thing ... I noticed that verbose=0 is that there's currently only the verbosity levels 0 and 1, and the verbosity can't be turned off. 

We could set `verbosity=-1` to turn the output off. But I would find it somewhat more intuitive to say

- `verbose=0` no output
- `verbose=1` (default) number of features in the current set
- `verbose=2` like above, plus time stamp and training/k-fold CV performance

I think we should also add it to the docstring then, and we could drop the `"avg."` from the output then I guess, since if `cv=0` it is just the single score, and k-fold CV score kind of implies "avg". What do you think?

wdm0006:
I went ahead and took a stab at those changes, but if you want to change the wording or anything like that, we can try out the new GitHub feature.

rasbt:
Looks good to me; I just checked it out, and I really like it, super useful. Thanks for the great contribution!

PR #111, opened 2016-09-29
adding estimator not fit exception

avloss:
without this exception it’s not clear how to work with ensemble
estimator - it seems that it’s impossible to fit each estimator
separately and then put them into ensemble - so there should be an
exception indicating that this estimator should be fit together!


rasbt:
Thanks for the PR, Anton. I haven't had a chance to look at it in detail, yet -- will get to that later today or tomorrow. However, I just noticed issue #112 . The mlxtend code should be compatible to both 0.17 and 0.18 now -- added that yesterday night (https://travis-ci.org/rasbt/mlxtend/builds/163611926). 

For adding the `NotFittedError` exception, I recommend using

```python
from distutils.version import LooseVersion as Version
from sklearn import __version__ as sklearn_version
if Version(sklearn_version) < '0.18':
    from sklearn.utils.validation import NotFittedError
else:
    from sklearn.exceptions import NotFittedError
```

for now. This can be removed in far future when most people made the switch (maybe sklearn 0.19 or 0.20 or so)

rasbt:
I am wondering, did you modify something in the documentation? I think if you could just make the change as suggested above (and maybe remove the documentation tweaks from the commit if you haven't changed anything), it should be fine and pass the integration tests for sklearn 0.18. Thanks!

avloss:
I didn't mean to modify anything. Let me fork from the right branch and recreate this pull request again!

avloss:
I will not add the check that you have suggested, it doesn't look right to me, I'll be committing to `rasbt:sklearn0.18` branch, so `sklearn` version `>=0.18` should be assumed

created new pull request https://github.com/rasbt/mlxtend/pull/120

rasbt:
> I will not add the check that you have suggested, it doesn't look right to me

Do you mean "technically" not right or style-wise? What do you think would be the better option to do this so that it doesn't crash on machine that haven't updated yet?

avloss:
sorry for posting so late. I definitely don't want any machines to crash. But it is my understanding that at some point mlxtend will have 0.18 as dependency, and I would like to commit into that branch! The change is not is valuable enough to clutter code with workarounds (this is obviously my humble opinion). It's super-minor, and it'll quietly wait it's turn, when .18 is a requirement! :) thanks again on great job on this library!

rasbt:
No worries at all :). 

> The change is not is valuable enough to clutter code with workarounds (this is obviously my humble opinion). It's super-minor, and it'll quietly wait it's turn, when .18 is a requirement! :)

I'd see this additional version-check clutter more as a temporary thing -- I think your PR just came in a few days after the sklearn 0.18 release. The old scikit-learn syntax still works in scikit-learn 0.18 (until 0.20 I think) but raises a deprecation warning to give people a chance to upgrade. However, using the new syntax in scklearn <= 0.17 will throw errors that may confuse people if they are not up to date yet (regarding scikit-learn's release). 

However, it think that making the most recent sklearn version a requirement is a reasonable thing to do (instead of cluttering up the code) for small open-source projects.

PR #126, opened 2016-11-15
refit param for votingclassifier

rasbt:
Adds  a `refit` param to  `EnsembleVoteClassifier` that prevents refitting classifiers if `refit=False` to save computational time.


StudioEtrange:
Maybe I am wrong, but even with refit=False
we have to call fit()
or self.clfs_ will be empty and troubles start :)

Even if we call fit we have to pass X and Y. Values that we might not have anymore (In case of pre fitted classifier, and saved on disk days before)

I have tried to pass fake X and Y, but now the problem is with the LabelEncoder which is always call

rasbt:
Hi, there,

thanks for the note and sorry about the confusion regarding the the`refit` parameter ... what I had in mind was that say s.o. already fit the classifiers that go into `self.clfs_` and doesn't want to refit them -- e.g., because it is computationally too expensive.

(I think how you interpreted it would offer a replacement for sth. like pickle?)

Now, when you do sth like

```
from sklearn import datasets

iris = datasets.load_iris()
X, y = iris.data[:, 1:3], iris.target

from sklearn import model_selection
from sklearn.linear_model import LogisticRegression
from sklearn.naive_bayes import GaussianNB 
from sklearn.ensemble import RandomForestClassifier
from mlxtend.classifier import EnsembleVoteClassifier
import numpy as np

clf1 = LogisticRegression(random_state=1)
clf2 = RandomForestClassifier(random_state=1)
clf3 = GaussianNB()

clf1.fit(X, y)
clf2.fit(X, y)
clf3.fit(X, y)

eclf = EnsembleVoteClassifier(clfs=[clf1, clf2, clf3], weights=[1,1,1], refit=False)
eclf.fit(X, y)
```

the `EnsembleVoteClassifier` won't create empty clones of `clf1`, `clf2`, `clf3` and refit them to `X y`, since that may be wasteful if that's been done before. One scenario where this would be particularly wasteful would be hyperparameter tuning of the `weights` since the `clf1`, `clf2`, `clf3` would be refit to the same training set every time if `refit=True` -- by setting `refit=False`, one can play around with the `weights` without refitting.

Hope that makes sense, and I'd be very happy about a PR to clarify the meaning of the `refit` param!

PS: For what you had in mind, there would be a somewhat "hacky" workaround, assigning classifiers to  `clfs` "manually"

```
eclf = EnsembleVoteClassifier(clfs=['void'], weights=[1,1,1], refit=False)

clf1.fit(X, y)
clf2.fit(X, y)
clf3.fit(X, y)

eclf.clfs = [clf1, clf2, clf3]

eclf.fit(X, y)
```

StudioEtrange:
Thanks
I understand why you did that use: For example : Replay weights as hyperparameters

Yes I have in mind some models saved with jobutil or pickle.
And I wanted to make a vote against them.
So the mechanism of "refit" could match my use.

So, i did a dirty hack :
if refit=False 
* in __init__ , self.clfs_ are initialized
* fit() become not necessary
* LabelEncoder is disabled

PR #137, opened 2016-12-22
Fix perfomance vs. training size plotting in plot_learning_curve

ipashchenko:
### Description

I found that picture in [user guide](http://rasbt.github.io/mlxtend/user_guide/plotting/plot_learning_curves/) with learning curve shows (i think) unexpected behaviour (higher misclassification error with larger training size).

### DIFF
--- a/mlxtend/plotting/learning_curves.py
+++ b/mlxtend/plotting/learning_curves.py
@@ -94,7 +94,7 @@ def plot_learning_curves(X_train, y_train,
     training_errors = []
     test_errors = []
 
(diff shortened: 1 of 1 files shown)


rasbt:
Good catch, thanks a lot. Btw. could you please create a new topic branch next time. I don't want to pull into the master branch directly -- in this case, it's such a simple fix so that's okay, but for more complex thinks that could cause all kinds of issues.

Thanks for the contribution

ipashchenko:
OK, thanks for useful tool!

Issue #156, opened 2017-03-08
Some plot titles seem to be inconsistent with plot parameters


pmarcelino:
In the Chapter 'Sequential Feature Selector' of the User Guide, some of the plot titles seem to be inconsistent with the plot parameters. 

For instance, in [Example 5](http://rasbt.github.io/mlxtend/user_guide/feature_selection/SequentialFeatureSelector/#example-5-sequential-feature-selection-for-regression), the plot parameter says _kind='**std_err**'_ but the plot title says '_(...) (w. **StdDev**)_'. 

I'm available to fix these small issues.

rasbt:
Good catch, thanks for the note. Sure, I'd welcome a PR to fix this little typo (you just need to change the typo in the Jupyter notebook). Otherwise, I'd be also happy to do it, since I need to push the changes to the website anyway :)

pmarcelino:
Great! If you're already pushing the changes to the website, maybe it's faster if you also fix the typo. I never contributed in a GitHub project, so I'd need to read how to do it first... :)

rasbt:
No worries, I just fixed the StdErr typo in #158 
However, please feel free to send in PR in future if you find sth. Open-source contributions can take some time in the beginning -- especially given how much work it is to go through the different steps for fixing a typo :P -- but especially typos are also a good way for getting acquainted with the mechanics :). Btw if it is useful in general, I have a quick walkthrough here: http://rasbt.github.io/mlxtend/contributing/#how-to-contribute and there's also a excellent "contributing" guide on scikit-learn: http://scikit-learn.org/stable/developers/contributing.html

PR #158, opened 2017-03-14
fix typo in sequential feature selection plot

rasbt:
Fixes typo in example 5 of the sequential feature selection notebook according to #156


rasbt:
Hm, interesting, there seems to be some issue with one of the unit tests on Windows now ...

PR #159, opened 2017-03-16
Rule generation

### LINKED ISSUE #157: Extend apriori function to also allow generation of rules
Hey Sebastian,

I am currently working on the topic of association analysis and after unsuccessfully looking through sklearn I found your package mlxtend - including `apriori.py` .

Based on your current implementation I do not see an option to generate rules from the list of frequent itemsets - or do I miss some other features beside the [apriori.py](https://github.com/bbardakk/mlxtend/blob/master/mlxtend/association/apriori.py) implementation?

I am not sure to what extent you or other contributors are already working on that topic. Seeing the issue #65 I was wondering if you already have included the rule generation, since - if I remember correctly - the discussed LIFT metric is based on the _support & confidence of association rules_.

If there is nothing in place I would be more than happy to help you implementing this feature based on the algorithm in _Tan, Pang-Ning, Michael Steinbach, and Vipin Kumar. Introduction to Data Mining. Pearson New International Edition.  Harlow: Pearson Education Ltd., 2014. (pp. 349-352)_.

Best
Joshua

JoshuaGoerner:
### Description

Implemented a rule generation algorithm to extract rules from frequent itemsets based on the metrics _confidece_, _lift_ <strike>and _conviction_</strike>.

Fixes #157 

### Pull Request requirements


rasbt:
Thanks for the PR, Joshua! Let's take the discussion from #157 over here then :).

I think we can rename the function `generate_rules` to `association_rules` since it would make it easier for someone to find it (since we are already renaming the `association` package to `frequent_patterns`). 

To be honest, I haven't implemented or used the "conviction" metric, yet, the number of "infs" in such a simple example is somewhat disturbing, though. So, I guess the denominator in that equation gets really, really small when the confidence gets really, really large. Right now, I have some doubts regarding the "conviction" metric (the general definition I found). It has (1 - confidence) in the denominator, and if the confidence score is 1.0, the metric would be undefined. Sorry, I suggested this metric because I stumbled upon it, but I am not really sure if it is useful. Maybe we should drop it from the table (and people can readily compute it from the DataFrame output if necessary). What do you think?

JoshuaGoerner:
> I think we can rename the function generate_rules to association_rules

Yes, makes it definitively easier to find

> I haven't implemented or used the "conviction" metric

Me neither :) _Support, Confidence & Lift_ have been sufficient for me. Let me drop the _conviction_ metric (this will also eliminate the `inf`s in the dataframe, even though they are calculated correctly).

Do you have other feedback w.r.t. the `association_rules()` function in mind? Otherwise I would focus on getting the tests & other cosmetics right.

rasbt:
The `association_rules` function looks pretty good! Just a minor comment I'll add to the code review...

mlxtend/association/association_rules.py
```diff
@@ -0,0 +1,61 @@
...
+        metric_dict["confidence"](sXY, sX, sY)/sY
+        }
+    
+    # check for metric compliance
+    if metric not in metric_dict.keys():
+        raise ValueError("Metric must be 'confidence' or 'lift', got '{}'".format(metric))
+        
+    # get dict of {frequent itemset} -> support
+    frozenset_vect = np.vectorize(lambda x: frozenset(x))
+    keys = df.values.T[1]
```
rasbt: I think using 

    keys = df['itemsets'].values
    values = df['support'].values

could be a safer approach. It would throw an error if the exact columns are not present in the input dataset rather than doing something unexpected where the code may run but produces weird results (e.g,. if someone swapped the column order or so)

mlxtend/association/association_rules.py
```diff
@@ -0,0 +1,61 @@
...
+                sY = frequent_items_dict[consequent]
+                # check for the threshold
+                if metric_dict[metric](sXY, sX, sY) >= min_threshold:
+                    rule_antecedants.append(antecedent)
+                    rule_consequents.append(consequent)
+                    rule_supports.append([sXY, sX, sY])
+                    
+    # check if frequent rule was generated
+    if not rule_supports:
+        return
```
rasbt: I'd say it's better to always return a DataFrame with the same column structure (makes unit testing and other things maybe easier). E.g., something like `pd.DataFrame(columns=["antecedants", "consequents", "support"] + list(metric_dict.keys()))`

mlxtend/association/association_rules.py
JoshuaGoerner: Yes

mlxtend/association/association_rules.py
JoshuaGoerner: yes

JoshuaGoerner:
Hey Sebastian,

I incorporated both of your valuable comments. In addition to that also 
- unit tests were written & executed
- same for coverage tests
- PEP8 style conformance was checked
- Feature was added to the changelog
- Documenting IPYNB was refactored

rasbt:
Thanks for the update! I will go through the code now and add a few comments here and there ... if I have any :)

rasbt:
- [ ] I made some small modifications to the Jupyter Notebook so that the frequent itemsets are generated in example 1, to make sure that the notebook executes okay. Also, I expanded a little bit on example 2 and added the API docs at the very bottom.  

If you don't mind, could you enable the `Allow edits from maintainers` feature so that I could add it to this PR (I thought it would make things easier). 
allow-maintainers-to-make-edits-sidebar-checkbox

mlxtend/association/__init__.py
```diff
@@ -1,10 +1,10 @@
-# Sebastian Raschka 2014-2017
+# Sebastian Raschka 2014-20
```
JoshuaGoerner: Strange, seems to be change by random...

JoshuaGoerner: Maybe I was a bit fast on changing modes in VIM 🙈

JoshuaGoerner:
Incorporated both code comments (header + typo in header). 

The [X]

mlxtend/association/association_rules.py
```diff
@@ -0,0 +1,93 @@
+from itertools import combinations
```
JoshuaGoerner: Done

rasbt:
Nice, thanks!

rasbt:
It looks pretty nice overall now! Thanks for the great contribution! 

For the long run, I think it'd be a good idea to rename `association` to `frequent_patterns` I think it's probably better to do this in a separate branch I guess -- before it gets more complicated and I screwn sth up with this PR :P. I think I will also make a new mlxtend release then so that we have the new functions and fixes and such available via pip, because they are pretty useful :)

JoshuaGoerner:
Very cool, will let you know when I stumble upon sth new to be included in mlxtend.

PR #190, opened 2017-05-13
adding hints for classifier predict methods to use per scoring function; addresses issue #176

### LINKED ISSUE #176: log_loss is calculated on class predictions, not probabilities
According to source code, currently `mlxtend.plotting.plot_learning_curves` always calls estimator's `predict` method which is not good approach for metrics like log loss.

It may be useful to use more high level abstraction like `sklearn.metrics.scorer.make_scorer` instead of metric itself.

P.S. It looks like docstring is a bit inaccurate: it contains "Plots learning curves of a classifier", but the function is suitable for regressors too, isn't it?

soumyadsanyal:
Fix for issue #176 

### Description

1. Uses the appropriate `predict` or `predict_proba` method for a classifier depending on which model scoring function is being examined via hints in the function definition.
2. Uses `make_scorer` to provide an interface for the scoring methods.
3. Potential TODO (not implemented in this PR): allow the caller to explicitly specify which of `predict`, `predict_proba`, `fit_and_predict` methods to use instead of doing the inference for it.

Also possibly addresses issue #192 

### Pull Request requirements


soumyadsanyal:
WIP.

soumyadsanyal:
@rasbt : this is a first pass at addressing issue #176 ; flipping over to you for review. Please let me know if any comments, corrections, suggestions, etc!

rasbt:
Looks really nice, thanks a lot! I see that you also addressed the issue with the unit tests raised in #192 . To be honest, I don't know why exactly the results deviate from the original values there. I suspect that I changed something a while ago (maybe a bugfix) and since the mlxtend.plotting package wasn't part of the unit tests on CI, it probably wasn't updated in the testing scripts. Everything looks correct to me know, and if there's no other concern from your side, I'd be happy to adjust the documentation and merge this (and re-enable mlxtend.plotting in the Travis CI tests so that sth like that doesn't happen again in future :)).

soumyadsanyal:
@rasbt sounds good to me! :)

rasbt:
Sorry, but on a second thought, I think we can simplify this greatly. E.g., there's a `check_scoring` function in scikit-learn that we could use to get the "right" `scorer` for a given metric/estimator automatically.

Also, maybe it would be good to rename  `clf` to `estimator`. Then we could add the following docstring for `scoring`:

```
"""
...
    scoring : str, callable, or None (default=None)
        Scoring-function based on scikit-learn:
        - None: Uses the estimator's default scoring function 
        - String (see http://scikit-learn.org/stable/modules/
                  model_evaluation.html#common-cases-predefined-values)
        - 'misclassification error': 1-accuracy
        - A scorer, callable object / function 
        with signature `scorer(estimator, X, y)`.
...
"""
```

In brief, the `check_scoring` could be used to create the `scorer` like so:

```python
    from sklearn.metrics.scorer import check_scoring
    scorer = check_scoring(estimator, scoring=scoring)
```

Here's a more detailed example demonstrating how this works:

```python
from sklearn.metrics.scorer import check_scoring
from sklearn import datasets
from sklearn.model_selection import train_test_split

iris = datasets.load_iris()
X = iris.data
y = iris.target
X_train, X_test, y_train, y_test = (train_test_split(X, y,
                                    train_size=0.6, random_state=2))

clf = LogisticRegression(random_state=1)
y_pred = clf.fit(X_train, y_train).predict(X_test)

scorer1 = check_scoring(clf, scoring=None)
score1 = scorer(clf, X_test, y_test)

scorer2 = check_scoring(clf, scoring='accuracy')
score2 = scorer(clf, X_test, y_test)

assert score1 == score2

print('Accuracy', score1)
```

```
Accuracy 0.966666666667
```


And for the misclassification error:

```python
if 'misclassification error':
    scorer3 = check_scoring(clf, scoring='accuracy')
    score3 = 1 - scorer(clf, X_test, y_test)
    print('misclassification error', score3)
```

Sorry about the "change of mind," but I think using `check_scoring` would make things more robust and "cleaner". What do you think?

soumyadsanyal:
@rasbt I definitely agree that would be better, I hadn't noticed the `check_scoring` method; great catch! I'd be happy to implement this in a couple of days if there's no rush to merge sooner than that.

rasbt:
Thanks, and there's definitely no rush :)

soumyadsanyal:
... sorry about the delay on this, was swamped with a few things. Will wrap this up shortly!

rasbt:
Thanks for the update, and again, no worries regarding the delay. Really appreciate the PR!

soumyadsanyal:
TODO: guards in complement_scorer for scorer methods evaluating within `[0,1]`. Also, tests.

rasbt:
Thanks for the update! 

>  maybe guards / try-catch in complement_scorer for scorer methods evaluating within [0,1]

I think that would be a good idea. Another thing, and it was a bad choice in hindsight, the default is "misclassification error" ... I'd say that's probably the most common use case for learning curves in literature, however, there would be a problem now if the estimator is a regressor. Maybe, we should change the default to `None`, which then uses the estimators default, and if the estimator does not have a default, it would raise a warning so that a user has to specify the metric. What do you think?

soumyadsanyal:
👍 Yep, that sounds good :). I'll update very shortly.

rasbt:
Thanks for the PR, I see that it is passing now :). 
Could you maybe include a regressor in the unit tests, just to make sure that it works as intended? 

PS: I am happy to take care of updating the documentation (the API doc, the Jupyter Notebook, and the Changelog) if you like (it would just require to check the "Allow edits from maintainers." option in the lower right corner)

PR #193, opened 2017-05-16
Multiprocessing over features rather than CV folds in Sequential Feature Selection (addressing #191)

### LINKED ISSUE #191: Sequential Feaute Selection CPU utilization
For multiple CPU utilization SFS uses [sklearn.model_selection.cross_val_score](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/model_selection/_validation.py#L137). It runs one process per validation fold. Usually there are more features than folds. So, it is more scalable to use one process per feature.

whalebot-helmsman:
### Description

Use 1 process per feature for sequential feature selection and exhaustive feature selection

Fixes #191

### DIFF
--- a/mlxtend/feature_selection/exhaustive_feature_selector.py
+++ b/mlxtend/feature_selection/exhaustive_feature_selector.py
@@ -20,6 +20,7 @@ from sklearn.base import BaseEstimator
 from sklearn.base import MetaEstimatorMixin
 from ..externals.name_estimators import _name_estimators
 from sklearn.model_selection import cross_val_score
+from joblib import Parallel, delayed
 
 
 class ExhaustiveFeatureSelector(BaseEstimator, MetaEstimatorMixin):
@@ -147,8 +148,13 @@ class ExhaustiveFeatureSelector(BaseEstimator, MetaEstimatorMixin):
 
         self.subsets_ = {}
         all_comb = len(candidates)
-        for iteration, c in enumerate(candidates):
-            c, cv_scores = self._calc_score(X=X, y=y, indices=c)
+        n_jobs = min(self.n_jobs, all_comb)
+        parallel = Parallel(n_jobs=n_jobs, pre_dispatch=self.pre_dispatch)
+
+
+        for iteration, (c, cv_scores) in enumerate(parallel(delayed(self._calc_score)
+                                                                        (X, y, c)
+                                                for c in candidates)):
 
             self.subsets_[iteration] = {'feature_idx': c,
                                         'cv_scores': cv_scores,
--- a/mlxtend/feature_selection/exhaustive_feature_selector.py
+++ b/mlxtend/feature_selection/exhaustive_feature_selector.py
@@ -66,7 +66,8 @@ class ExhaustiveFeatureSelector(BaseEstimator, MetaEstimatorMixin):
         otherwise.
         No cross-validation if cv is None, False, or 0.
     n_jobs : int (default: 1)
-        The number of CPUs to use for cross validation. -1 means 'all CPUs'.
+        The number of CPUs to use for evaluating different feature subsets
+        in parallel. -1 means 'all CPUs'.
     pre_dispatch : int, or string (default: '2*n_jobs')
         Controls the number of jobs that get dispatched
         during parallel execution in cross_val_score.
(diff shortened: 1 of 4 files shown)


rasbt:
Looks great, thanks a lot! I will add a few minor code comments regarding the documentation. (Alt. I am happy to add those to this PR if you do the "enabling repository maintainer permissions on existing pull requests" feature.)

rasbt:
Okay, it didn't let me add the comments directly to the code since it wasn't modified by you in this PR, so let me post it here :)

Could you change the docstrings for `n_jobs` and `pre_dispatch` slightly to the following (since they are currently referencing the old cv_score parallelization):

```
"""
...
    n_jobs : int (default: 1)
        The number of CPUs to use for evaluating different feature subsets
        in parallel. -1 means 'all CPUs'.
    pre_dispatch : int, or string (default: '2*n_jobs')
        Controls the number of jobs that get dispatched
        during parallel execution if `n_jobs > 1` or `n_jobs=-1`.
        Reducing this number can be useful to avoid an explosion of
        memory consumption when more jobs get dispatched than CPUs can process.
        This parameter can be:
        None, in which case all the jobs are immediately created and spawned.
            Use this for lightweight and fast-running jobs,
            to avoid delays due to on-demand spawning of the jobs
        An int, giving the exact number of total jobs that are spawned
        A string, giving an expression as a function
            of n_jobs, as in `2*n_jobs`
...
"""
```

Also, could you add the following snippet to the Changelog in docs/sources/CHANGELOG.md?

```
##### Changes

...

- Parallel execution in `mlxtend.feature_selection.SequentialFeatureSelector` and `mlxtend.feature_selection.ExhaustiveFeatureSelector` is now performed over different feature subsets instead of the different cross-validation folds to better utilize machines with multiple processors if the number of features is large ([#193](https://github.com/rasbt/mlxtend/pull/193), via [@whalebot-helmsman](https://github.com/whalebot-helmsman)).

```

Otherwise, the PR looks great, thanks a lot!

mlxtend/feature_selection/exhaustive_feature_selector.py
```diff
@@ -66,7 +66,8 @@ class ExhaustiveFeatureSelector(BaseEstimator, MetaEstimatorMixin):
         otherwise.
         No cross-validation if cv is None, False, or 0.
     n_jobs : int (default: 1)
-        The number of CPUs to use for cross validation. -1 means 'all CPUs'.
+        The number of CPUs to use for evaluating different feature subsets
+        in parallel. -1 means 'all CPUs'.
     pre_dispatch : int, or string (default: '2*n_jobs')
         Controls the number of jobs that get dispatched
         during parallel execution in cross_val_score.
```
rasbt: Thanks! Could you also update the pre_dispatch docstring since it still has the cross_val_score referenced there, which may be misleading

mlxtend/feature_selection/sequential_feature_selector.py
```diff
@@ -84,7 +84,8 @@ class SequentialFeatureSelector(BaseEstimator, MetaEstimatorMixin):
         exclusion/inclusion if floating=True and
         algorithm gets stuck in cycles.
     n_jobs : int (default: 1)
-        The number of CPUs to use for cross validation. -1 means 'all CPUs'.
+        The number of CPUs to use for evaluating different feature subsets
+        in parallel. -1 means 'all CPUs'.
     pre_dispatch : int, or string (default: '2*n_jobs')
         Controls the number of jobs that get dispatched
         during parallel execution in cross_val_score.
```
rasbt: also here. thanks!

whalebot-helmsman:
Added changelog and documentation updates

rasbt:
I have to look into Travis CI  to see if it supports multiple processors to better tests PRs like that. However, I just tried it locally on 8 cpus and it seems to yield the same results as in the unit tests with 1 processor -- everything seems to work as expected and I will merge it now. 

Thanks a lot for this really useful contribution!

whalebot-helmsman:
Do you use unit-tests for comparison? There are very easy (in a sense of CPU time required) tasks in unit-tests. In such scenario overhead of starting new processes is bigger than gains of multiprocessing.
[test.py](https://gist.githubusercontent.com/whalebot-helmsman/28b875cc0edb2bd6f69bffbad3a02162/raw/e5102927217900bc730e91d2217ed17bdf334f27/test.py) imitates hard (in a sense of CPU time required) train sets with slow classifier on a small public dataset. You can measure it (parameter is n_jobs)

```
time python test.py 1
time python test.py 2
time python test.py 4
time python test.py 8 
```

rasbt:
> Do you use unit-tests for comparison? 

Yeah, I was primarily looking for correctness. I.e., that the previous results in the unit tests are reproduced with > 1 CPUs.

Issue #194, opened 2017-05-20
Powerset generator in EFS consumed 240 GB of RAM for ~ 300 features


adam-erickson:
The _candidates_ object in _EFS.fit_ implements a standard powerset generator using _itertools_. This function quickly consumed 240 GB of RAM with ~300 features for the range (0, 20). This makes sense given that this generates around 7.5^30 8-byte integers, making it too large for memory even in a cluster environment.

I'm currently testing a solution utilizing _joblib_ _delayed_ and _Parallel_ along with _dask.array_, which looks like this:

```
import dask.array as da

def powerset(iterable, min_features, max_features)
    """
    returns a generator following the form:
    powerset([1,2,3] for range [1,3]) --> (1,) (2,) (3,) (1,2) (1,3) (2,3) (1,2,3)
    """
    s = list(iterable)
    return chain.from_iterable(combinations(s, r) for r in range(min_features, max_features + 1))

min_features, max_features = [1, 20]
ps = powerset(iterable=X, min_features=min_features, max_features)
def ret(x):
    return x
ps = [Parallel(n_jobs=4)(delayed(ret)(i) for i in ps)]
ps = da.from_delayed(ps)
```

So far, memory consumption appears dramatically reduced, but CPU utilization is 100% on a single core.

Cheers,

Adam

rasbt:
Thanks for the feedback. I never thought someone would actually run this exhaustive feature selection on such a large set of features, but yeah, why not fixing it so that it at least theoretically works :).

So, instead of generating the candidates upfront, we could use sth similar to your solution above (which is really elegant btw!) or a simple generator like so:

```python
from itertools import combinations


def powerset(min_features, max_features):
    for ln in range(min_features, max_features+1):
        for feature_set in combinations(range(max_features), r=ln):
            if feature_set:
                yield feature_set
    
            
min_features = 3
max_features = 8

for candidate in powerset(min_features, max_features):
    # evaluate
    pass
```

I can take a look at, or if you'd like to submit a PR, that'd be nice :)

adam-erickson:
PR added. The trouble with the powerset function you posted above is that it doesn't take into account the number of possible items to draw from, but is fixed as equal to max_features. The function I originally posted above would have required a list of column names or indices, which is also slightly less efficient.

rasbt:
Fixed via #195

PR #195, opened 2017-05-21
Modified powerset function in EFS to reduce memory consumption

adam-erickson:
### Description

I modified the powerset function and calculation of number of possible combinations to reduce memory overhead by not longer storing all combinations as a list.

Fixes # https://github.com/rasbt/mlxtend/issues/194

### Pull Request requirements

### DIFF
--- a/mlxtend/feature_selection/exhaustive_feature_selector.py
+++ b/mlxtend/feature_selection/exhaustive_feature_selector.py
@@ -166,6 +166,20 @@ class ExhaustiveFeatureSelector(BaseEstimator, MetaEstimatorMixin):
         self.subsets_ = {}
         
         def ncr(n, r):
+        """Return the number of combinations of length r from n items.
+        
+        Parameters
+        ----------
+        n : {integer}
+            Total number of items
+        r : {integer}
+            Number of items to select from n
+            
+        Returns
+        -------
+        Number of combinations, integer
+        
+        """
             r = min(r, n-r)
             if r == 0:
                 return 1
--- a/mlxtend/feature_selection/exhaustive_feature_selector.py
+++ b/mlxtend/feature_selection/exhaustive_feature_selector.py
@@ -166,20 +166,21 @@ class ExhaustiveFeatureSelector(BaseEstimator, MetaEstimatorMixin):
         self.subsets_ = {}
         
         def ncr(n, r):
-        """Return the number of combinations of length r from n items.
-        
-        Parameters
-        ----------
-        n : {integer}
+            """Return the number of combinations of length r from n items.
+            
+            Parameters
+            ----------
+            n : {integer}
             Total number of items
-        r : {integer}
+            r : {integer}
             Number of items to select from n
             
-        Returns
-        -------
-        Number of combinations, integer
-        
-        """
+            Returns
+            -------
+            Number of combinations, integer
+            
+            """
+            
             r = min(r, n-r)
             if r == 0:
                 return 1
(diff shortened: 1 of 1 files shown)


rasbt:
Thanks for the PR! I just gave it a quick glance and there's already something like `ncr` ([`math.num_combinations`](https://github.com/rasbt/mlxtend/blob/master/mlxtend/math/counting.py#L17)); prob a bit more efficient since it doesn't evaluate range objects.

E.g., you could replace 

```python
all_comb = np.sum([ncr(n=X.shape[1], r=i)
                   for i in range(self.min_features,
                                  self.max_features + 1)])
```

by

``python
from ..math import num_combinations
...
all_comb = np.sum([num_combinations(n=X.shape[1], k=i)
                   for i in range(self.min_features,
                                  self.max_features + 1)])
```

adam-erickson:
Supposedly, the implementation provided is ~25% more efficient than math.factorial: http://stackoverflow.com/questions/4941753/is-there-a-math-ncr-function-in-python

rasbt:
Interesting! However, I think any of these functions run so quickly that speed is really not a concern here. Sorry my previous comment was a bit misleading; I primarily thought that resusing the existing num_combinations function would be a good idea to avoid reimplementing this code here (in the sense of refactoring or duplication of effort :)).

rasbt:
Hi, @adam-erickson . Just revisiting this discussion from last weekend ...
That's a really good fix. and I'd like to merge it. Can we use the internal `from ..math import num_combinations` for this though to keep the code simple and add a short message to the Changelog? I am happy to make those changes via "Allow edits from maintainers" if you don't mind :)

adam-erickson:
Hi @rasbt sorry for the slow reply, I was releasing [gapfraction](https://github.com/adam-erickson/gapfraction) and its website. Of course, we can use the _num_combinations_ function if you'd like for simplicity. What can I do to help facilitate these changes? I'm a bit new to this process.

rasbt:
No worries at all -- just wasn't sure if you'd okay with the proposed changes. 

> What can I do to help facilitate these changes? I'm a bit new to this process. Should I close the pull request?

Feel free to make another commit to this PR to change the code using the existing 

    from ..math import num_combinations

and add a note to the CHANGELOG.md file at ./docs/sources/CHANGELOG.md (in the bug fixes section).

Since GitHub now has this convenient "Squash and Merge" button, we don't have to worry about "too many" commits since that's all being taken care of at the end.

adam-erickson:
Hi @rabst, I did some testing and it appears that the results of the `num_combinations` function may be imprecise or incorrect? I tested it against the `scipy.misc.comb` function with `exact=True` for `N=300, k=10` and found different outputs. The proposed `nCr `function produces identical results to `scipy.misc.comb`. You might want to verify this.

rasbt:
Thanks for looking into that, @adam-erickson . I also find the same issue, when I 

    >>> num_combinations(n=200, k=10)
    22451004309013280

   >>> scipy.misc.comb(n=200, k=10)
   22451004309013280

But for n=300, I get `1398320233241701888` & `1398320233241701770` (the later from scipy). I ran the same thing in Python 2.7 and noticed that Py 2.7 gives the correct results for `num_combinations`. Turns out to be an integer division problem. For instance, it should be 

    comb = int(numerator//denominator)

instead of 

    comb = int(numerator/denominator)

rasbt:
Should be fixed now in #200 (but you could also use scipy.misc (didn't know `comb` existed back then)

rasbt:
Sorry, somehow, I forgot about this ... Merging now because the code is good and works perfectly fine :). Thanks for the contribution!

PR #203, opened 2017-06-14
Support for abstract cv method for StackingCVClassifier

sque:
### Description
Change the API of `StackingCVClassifier` in order to accept any kind of cross-validation technique. 

### Related issues or pull requests

Related #202

### Pull Request requirements

- [x] Added appropriate unit test functions in the `./mlxtend/*/tests` directories
- [x] Ran `nosetests ./mlxtend -sv` and make sure that all unit tests pass
- [x] Checked the test coverage by running `nosetests ./mlxtend --with-coverage`
- [x] Checked for style issues by running `flake8 ./mlxtend`
- [x] Added a note about the modification or contribution to the `./docs/sources/CHANGELOG.md` file
- [ ] Modify documentation in the appropriate location under `mlxtend/docs/sources/` (optional)
- [ ] Checked that the Travis-CI build passed at https://travis-ci.org/rasbt/mlxtend

PR #204, opened 2017-06-15
Support for abstract cv method for StackingCVRegressor

rasbt:
Similar to #203 but for `StackingCVRegressor` instead of `StackingCVClassifier`


rasbt:
Hi, @EikeDehling Just wanted to cc you on this in case you are interested. Just a few minor changes to accept any scikit- CV object

PR #249, opened 2017-09-18
Inheritance in OnehotTransactions

iaroslav-ai:
The following
```python
from sklearn.base import clone
from mlxtend.preprocessing import OnehotTransactions
enc = OnehotTransactions()
clone(enc)
```
raises
```
TypeError: Cannot clone object '<mlxtend.preprocessing.onehot.OnehotTransactions object at 0x7f11d8ed7a90>' (type <class 'mlxtend.preprocessing.onehot.OnehotTransactions'>): it does not seem to be a scikit-learn estimator as it does not implement a 'get_params' methods.
```
because the class does not inherit from base classes in sklearn.
This is a fix


iaroslav-ai:
P.S. Because of this I cannot use this class for example in FeatureUnion or in GridSearchCV.

rasbt:
Yeah, that makes sense. Thanks a lot!

PR #262, opened 2017-10-12
recursive floating for SequentialFeatureSelector

### LINKED ISSUE #243: SFFS & SFBS recursively conditionally removing/adding features
Hey,

(Will talk about SFFS)

I went through the code and I think there is something missing.

image

So, basically, you are missing a while inside the while for Step 3. The algorithm should continue conditionally removing features until either you cannot (cause condition doesn't hold) or only 2 features are remaining.

rasbt:
### Description

Adds an recursive_floating parameter to the `SequentialFeatureSelector`:

    recursive_floating : bool (default=False)
        If `True`, uses the floating behavior described in
        Novovicova & Kittler (1994) for the `floating=True` variants.
        That is, the continuation of conditional exclusion in SFFS
        (skipping inclusion if feature removal leads to better performance).
        Similarly, the conditional inclusion is continued in SFBS
        (and consequently the the exclusion is skipped) as long as it leads
        to improvements in performance.
        Note: `recursive_floating` will be set `True` in future versions
        of mlxtend (mlxtend > 0.10).

Fixes #243 

### Pull Request requirements

More todos:

- [ ] add a deprecation warning to the default setting `recursive_floating=False` (only if `floating=True` to avoid being too annoying) mentioning that this will be  `recursive_floating=True` in mlxtend >= v0.10


rasbt:
I simplified the implementation a little bit using two nested while-loops now. This is doing exactly the same as before with `recursive_floating=True` based on my debugger, but is now hopefully a bit more intuitive to read. Note that I removed the `recursive_floating` parameter and enabled the implementation by default now as the implementation is already complicated enough :P (also I didn't notice any differences in the test scenarios; I believe it could improve the feature selection in certain edge cases though)

rasbt:
just saw that we had the discussion in the other thread, so I am just pinging you here ;) @nd26

PR #272, opened 2017-10-31
Extend functionality of Sequential Feature Selector to allow repeating cross-validation.

### LINKED ISSUE #271: Add "RepeatedStratifiedKFold" (from sklearn) on top of cross-validation in sequential feature selectors
As reported by @nd26, adding "RepeatedStratifiedKFold" (i.e., repeating cv a user-specified number of times) may yield to better results on certain datasets. This could be added as an additional parameter to the `SequentialFeatureSelector`, for example, `reapeat_cv`, which should default to 0 and accept positive integers

nd26:
### Description

Added ‘n_cv_repeats’ parameter and extended functionality with RepeatedStratifiedKFold and RepeatedKFold to allow repeating cross-validation for both classifiers and regressors respectively.

Fixes #271 

### Pull Request requirements

### DIFF
--- a/mlxtend/feature_selection/sequential_feature_selector.py
+++ b/mlxtend/feature_selection/sequential_feature_selector.py
@@ -18,15 +18,19 @@ from sklearn.base import clone
 from sklearn.base import BaseEstimator
 from sklearn.base import MetaEstimatorMixin
 from ..externals.name_estimators import _name_estimators
-from sklearn.model_selection import cross_val_score
+from sklearn.model_selection import cross_val_score, RepeatedKFold, RepeatedStratifiedKFold
 from sklearn.externals.joblib import Parallel, delayed
-
 
 def _calc_score(selector, X, y, indices):
     if selector.cv:
+        if selector.n_cv_repeats > 0:
+            if selector._estimator_type == 'classifier':
+                cv_folds = RepeatedStratifiedKFold(n_splits=selector.cv, n_repeats=selector.n_cv_repeats)
+            else:
+                cv_folds = RepeatedKFold(n_splits=selector.cv, n_repeats=selector.n_cv_repeats)  
         scores = cross_val_score(selector.est_,
                                  X[:, indices], y,
-                                 cv=selector.cv,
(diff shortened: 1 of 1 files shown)


rasbt:
Thanks for the PR! Based on the test results, it just seems to be a Tab vs. Whitespace error. I.e., try to replace tabs by 4 whitespaces each.

Btw. you can always test your code locally by running `nosetests mlxtend/feature_selection/`, for example.

nd26:
No problem! Thanks for the reply. Will do tomorrow morning :)

rasbt:
Sure, no hurry!

Btw. added an issue regarding the random number generator you mentioned via email (#274). I am not sure what's more convenient, adding the Repeated CV feature or the random seed first. But I think adding repeated CV (using the current cross_val_score implementation) should work without an extra random seed. After that is implemented, I am happy to rewrite the code to support random seeds. 

However, I just wanted to mention it in case the repeated CV causes troubles without random seeds (i.e., irreproducible unit tests). In that case, we may want to add random seeds first and then the repeated CV. 

Anyways, thanks a lot for contributing, both repeated CV and the random seeds are useful enhancements!

nd26:
Most likely, if there are unit tests that test the n_cv_repeats parameter, they will fail due to the lack of reproducibility (SFS is highly unstable when used against a large number of features). But, current tests only test cross_val_score, right?

No problem man!

rasbt:
I just see the reason why one of the unit tests is failing is because one of them is running sklearn 0.18, which doesn't yet have the `RepeatedKFold` class (it was added in sklearn 0.19).

I think it's fair to switch all the unit tests to sklearn >= 0.19 as it has been out now for a while. 

The way to fix this would be to change `SKLEARN_VERSION="0.18"` in the `.travis.yml` file to `SKLEARN_VERSION="0.19"` in the section

```
- python: 3.5
          env: LATEST="false" COVERAGE="false" NUMPY_VERSION="1.11.2" SCIPY_VERSION="0.18.1" SKLEARN_VERSION="0.18" PANDAS_VERSION="0.19.1"
```

PR #278, opened 2017-11-03
Added optional name field to utils.counter

matsavage:
### Description

Added a naming variable to utils.counter and chnaged time precision

Link related issues/pull requests here

### Pull Request requirements

### DIFF
--- a/mlxtend/utils/counter.py
+++ b/mlxtend/utils/counter.py
@@ -45,7 +45,7 @@ class Counter(object):
         """Print current iteration and time elapsed."""
         self.curr_iter += 1
         out = '%d iter | %.2f sec' % (self.curr_iter, time.time() -
-                                    self.start_time)
+                                      self.start_time)
         if self.name is None:
             self.stream.write('\r%s' % out)
         else:
--- a/mlxtend/utils/counter.py
+++ b/mlxtend/utils/counter.py
@@ -13,7 +13,6 @@ import sys
 class Counter(object):
 
     """Class to display the progress of for-loop iterators.
-
     Parameters
     ----------
     stderr : bool (default: True)
@@ -21,25 +20,33 @@ class Counter(object):
     start_newline : bool (default: True)
         Prepends a new line to the counter, which prevents overwriting counters
         if multiple counters are printed in succession.
+    precision: int (default: 0)
+        Sets the precison of the displayed iteration time.
     name : string (default: None)
         Prepends the specified name before the counter to allow distinguishing
         between multiple counters.
-
     Attributes
     ----------
     curr_iter : int
         The current iteration.
     start_time : int
         The system's time in seconds when the Counter was initialized.
-
+    iteration_time : int
+        The system's time in seconds when the Counter was last updated.
     """
-    def __init__(self, stderr=False, start_newline=True, name=None):
+    def __init__(self, stderr=False, start_newline=True, precision=0,
+                 name=None):
         if stderr:
             self.stream = sys.stderr
         else:
             self.stream = sys.stdout
+        if isinstance(precision, int):
+            self.precision = '%%.%df' % precision
+        else:
+            self.precision = '%d'
         self.name = name
         self.start_time = time.time()
+        self.iteration_time = time.time()
         self.curr_iter = 0
         if start_newline:
             self.stream.write('\n')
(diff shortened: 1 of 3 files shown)


rasbt:
Thanks for the PR. How about adding a `precision_format` parameter instead of changing the default behavior (or do you have a better name than `precision_format` in mind?)

E.g., by default, this would be 

`precision_format=%d`, and it could then be flexibly changed to any desired precision like `precision_format=%.2f` etc.

matsavage:
@rasbt Great idea! 

I added a kwarg that accepts an int, for the number of decimal places to display, which seems cleaner.

To update the precision string I had to do a little refactoring, so added some methods to return the total time elapsed and the time elapsed in the current iteration.

rasbt:
Sounds good! Regarding the `iteration_time`, I don't think that's currently needed in your implementation though as I don't see it currently used anywhere except `iteration_elapsed`?

matsavage:
It's not used internally, but it seems like a useful thing to have, and goes along with the iteration_count variable

rasbt:
Thanks for the PR, I really appreciate that! However, I'd prefer to keep the Counter class relatively lean and simple. I.e., not adding not too many bells and whistles (methods) for one-liners.

Instead, I would suggest to add the desired features as an example to the documentation, e.g., see the "Examples" section in the docsctring below:


```python

class Counter(object):

    """Class to display the progress of for-loop iterators.

    Parameters
    ----------
    stderr : bool (default: True)
        Prints output to sys.stderr if True; uses sys.stdout otherwise.
    start_newline : bool (default: True)
        Prepends a new line to the counter, which prevents overwriting counters
        if multiple counters are printed in succession.
    precision: int (default: 0)
        Sets the precison of the displayed iteration time.
    name : string (default: None)
        Prepends the specified name before the counter to allow distinguishing
        between multiple counters.

    Attributes
    ----------
    curr_iter : int
        The current iteration.
    start_time : float
        The system's time in seconds when the Counter was initialized.
    end_time : float
        The system's time in seconds when the Counter was last updated.

    Examples
    --------
    >>> cnt = Counter()
    >>> for i in range(20):
    ...     # do some computation
    ...     time.sleep(0.1)
    ...     cnt.update()

    20 iter | 2 sec
    >>> print('The counter was initialized.'
              ' %d seconds ago.' % (time.time() - cnt.start_time))
    The counter was initialized 2 seconds ago
    >>> print('The counter was last updated'
              ' %d seconds ago.' % (time.time() - cnt.end_time))
    The counter was last updated 0 seconds ago.

    """
    def __init__(self, stderr=False, start_newline=True, precision=0,
                 name=None):
        if stderr:
            self.stream = sys.stderr
        else:
            self.stream = sys.stdout
        if isinstance(precision, int):
            self.precision = '%%.%df' % precision
        else:
            self.precision = '%d'
        self.name = name
        self.start_time = time.time()
        self.curr_iter = 0
        if start_newline:
            self.stream.write('\n')

    def update(self):
        """Print current iteration and time elapsed."""
        self.curr_iter += 1

        self.end_time = time.time()

        out = '%d iter | %s sec' % (self.curr_iter,
                                    self.precision % (self.end_time
                                                      - self.start_time))
        if self.name is None:
            self.stream.write('\r%s' % out)
        else:
            self.stream.write('\r %s: %s' % (self.name, out))
        self.stream.flush()
```

matsavage:
@rasbt That makes much more sense, I got the idea of wanting to keep things lean from the class as it currently stands, which was why I didn't include the precision kwarg on my first iteration.

Good idea on including that as an example, much simpler than my implementation. Sorry about so much back and forth about such a simple change, I'm new to github and contributing changes.

rasbt:
No worries, we've been all new to GitHub at some point :). Let me make a few more inline comments ...

mlxtend/utils/counter.py
```diff
@@ -21,20 +21,49 @@ class Counter(object):
     start_newline : bool (default: True)
         Prepends a new line to the counter, which prevents overwriting counters
         if multiple counters are printed in succession.
+    precision: int (default: 0)
+        Sets the precison of the displayed iteration time.
```
rasbt: There's a little typo and it should be "precision" not "precison". However, I suggest changing the short description altogether ... E.g., the number 1234567.89 has a precision of 9, so it would be more "precise" (pun intended :P) to refer to it as:

    Sets the number of decimal places when displaying the time elapsed in seconds.

matsavage: Haha, very good, change made

rasbt:
Looks good now, thanks! Will merge it and take care of the docs.

PR #279, opened 2017-11-05
Update documentation of the Counter docs

rasbt:


matsavage:
Thanks! I didn't know about this file at all

rasbt:
No worries! I didn't want to overwhelm with too many things at once. The documentation is build from the Jupyter notebooks which are in the docs/sources/user_guides subdirectory. If there's an API change, the `make_api.py` scripts needs to be run first as it generates the markdown files of the docstrings that get then included in the jupyter notebooks (at the very bottom). Then, after the Jupyter notebook is run, the script `ipynb2markdown.py` is run as `ipynb2markdown.py --ipynb path_to_the_updated_notebook` to generate the updated markdown file, which gets later converted to HTML/the web documentation when a new mlxtend version is released.

PR #280, opened 2017-11-05
Add "fuzzy" option to forward sequential feature selector

bschreck:
### Description

Adds two parameters to SFS- fuzzy and threshold. Fuzzy is a boolean, and threshold is a float. The idea is to add a separate, more efficient randomized algorithm to sequential feature selection which aids in feature selection on large datasets, datasets with many features, or pipelines that take a long time to compute. Instead of exhaustively searching through every feature on every iteration to find the one that _most_ improves the score, this algorithm instead randomly samples features to add until it finds one that improves the score by some threshold.

If need be, I can provide examples on public datasets showing this working and why there is a use case for it. In particular, I am using the open source [featuretools](https://github.com/featuretools/featuretools) library to automatically generate thousands of features for a dataset. This randomized feature selector works particularly well for pruning _wide_
feature matrices (more features than instances) .

I started working on updating the User Guide with the algorithm, but my math formatting is off, and I'm not sure the best way to add it. It works better as pseudocode than a mathematical equation. I also have not yet added an example to the code section of that page.


rasbt:
Thanks for the PR, Ben. I am currently traveling and can't check it out in detail. But based on your description, it sounds like an awesome idea & contribution. However, I think the `SequentialFeatureSelector` is already quite complex in terms of usage (i.e., there are already so many parameters and 4 variants: forward, backward, floating forward, floating backward).

So, instead of adding those additional variants to the existing implementation, I was thinking if it would make sense to make a new class for that. I.e., `RandomizedSFS` or `FuzzySFS` -- or maybe a completely different name if you have something better in mind :). This would maybe help users to better understand what's going on under the hood (also with respect to the formatting user documentation).

What do you think?

bschreck:
Yeah I think that's a great idea. It's a different enough concept too that it's not really "sequential" anymore. I'll update and push

rasbt:
> Yeah I think that's a great idea. It's a different enough concept too that it's not really "sequential" anymore. I'll update and push

Sounds like a good plan! :)

sashml:
>I am using the open source featuretools library to automatically generate thousands of features for a dataset. 

By the way, that approach is way toward *auto feature engineering*. And thanks for mentioning/sharing the library!

rasbt:
Yeah, that's what I suspected based on the code setup. Actually, I had randomized feature selection on my to do list (as an analog to randomized search -- versus grid search) based on the ideas on Bengio's paper. Anyways, I'd welcome any PR on this!

PR #327, opened 2018-02-09
Apriori/one hot speedup

jaksmid:
On my computer, the Apriori algorithm could not finish in reasonable time with the document containing millions of rows.
I optimized the code so the algorithm runs much faster

* Transform: One hot is much faster
* Apriori is much faster - some invalid combinations omitted


rasbt:
This is great, thanks a lot! 

Yeah, omitting items/itemsets without enough support in the previous step actually makes sense.

Not sure, but do you have "allow maintainer edits" enabled on this PR? (I wish GitHub would add a visual clue for that). I wanted to adjust the documentation a bit for consistency.

jaksmid:
Hello @rasbt,
glad that you like my PR.

I see  `allow edits from maintainers` as checked. But if you send me what you want to have changed I will be happy to incorporate the amendments for you.

rasbt:
Thanks! There are actually only two little things that come to mind:

- it's not included in the API doc because it's an internal function, but could you change the docstring of "generate_new_combinations" to the NumPy-style docstring format for consistency with other docstrings?

- An entry in the Changelog (mlxtend/docs/sources/CHANGELOG.md) would be nice. I just see that there's no category for "improvements" but I think the "Changes" category should suffice :)

jaksmid:
Done. Is this something you had in mind @rasbt?

rasbt:
Thanks for the update and the PR. Looks good to me now and I am happy to merge this :)

Issue #298, opened 2017-12-04
Supporting out-of-core processing in the apriori function for DataFrames that don't fit into memory


fernandocamargoai:
According to the docs, I should call `apriori(df, min_support=0.6, use_colnames=True)`. But what if the dataframe doesn't fit into the memory? Isn't there anyway I could submit smaller dataframes and aggregate the results?

rasbt:
Right now, there's no way to do that with regard to how it's implemented. However, that's actually a good point, and I think it would be a useful feature to submit smaller dataframes as batches. Regarding the implementation of such a "feature," maybe Dask could be used for this? I.e., having a parameter like `dataframe_backend='pandas'` (default) and having `dataframe_backend='dask'` to allow out-of-core processing. What do you think?

fernandocamargoti:
Well, I guess it's a good idea. I haven't used Dask yet, But I had a look at it and it seems promissing.

jaksmid:
When I was optimising the Apriori code (see https://github.com/rasbt/mlxtend/pull/327), I usually converted the dataframe into sparse one. It should not be a problem to build the sparse matrix beforehand. Especially if you have lots of products and small orders this could be able to handle much more data.

If/After the pull request is merged I may adjust the code for the sparse matrices

jaksmid:
So I have the implementation for the onehot transform that uses the sparse format.

Then all you have to do is use the `SparseDataFrame` instead of a regular one. 
Should I prepare the pull request?

rasbt:
Regarding the SparseDataFrame suggestion, that's a great idea to improve memory consumption with requiring any extra dependencies for the time being. Would really appreciate a PR

jaksmid:
Will prepare it in a day or so

rasbt:
@fernandocamargoti  while it's not out-of-core learning, you may find the PRs #328 & #329 helpful which are now in the master branch (thanks to @jaksmid ) (have a look at example 3 here: https://github.com/rasbt/mlxtend/blob/master/docs/sources/user_guide/frequent_patterns/apriori.ipynb)

In my tests on a toydataset with 100000 entries, the memory consumption shrank from 4.5 Gb to 600 Mb, and maybe that's helpful for your particular use case, for now ... until I find more time in future to support out of core fitting.

PR #328, opened 2018-02-14
Added sparse option to onehot transform

jaksmid:
Added option to create sparse one hot matrices on `onehot transform`

Should mitigate issue [#298](https://github.com/rasbt/mlxtend/issues/298)


jaksmid:
Hello @rasbt , please check if the API changes are to your liking. 

Wasnt sure about two possible format outcomes in the `transform return` doc.

Also not sure about the documentation notebooks though. Should I change them?

mlxtend/preprocessing/onehot.py
```diff
@@ -119,12 +120,19 @@ def transform(self, X):
...
+
+        sparse: bool
+          If True, transform will return Compressed Sparse Row matrix
+          instead of the regular one.
+
         Returns
         ------------
-        onehot : NumPy array [n_transactions, n_unique_items]
-           The NumPy one-hot encoded integer array of the input transactions,
+        onehot : NumPy array [n_transactions, n_unique_items] If not sparse
```
rasbt: I would suggest changing 

"If not sparse" 

to 

"if `sparse=False` (default)   
and  SciPy Compressed Sparse Row (CSR) matrix, otherwise

mlxtend/preprocessing/onehot.py
```diff
@@ -119,12 +120,19 @@ def transform(self, X):
            ['Milk', 'Beer', 'Rice'],
            ['Milk', 'Beer'],
            ['Apple', 'Bananas']]
+
+        sparse: bool
```
rasbt: suggested change

sparse: bool -> sparse: bool (default=False)

mlxtend/preprocessing/tests/test_onehot_transactions.py
```diff
@@ -47,13 +47,19 @@ def test_fit():
...
     oht.fit(dataset)
     trans = oht.transform(dataset)
     np.testing.assert_array_equal(expect, trans)
 
 
+def test_transform_sparse():
+    oht = OnehotTransactions()
+    oht.fit(dataset)
+    trans = oht.transform(dataset, sparse=True)
+    np.testing.assert_array_equal(expect, trans.todense())
```
rasbt: here, for readability, adding a line 

    assert (isinstance(trans, csr_matrix))

wouldn't hurt :)

docs/sources/CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ The CHANGELOG for the current development version is available at
 - Raises an informative error message if `predict` or `predict_meta_features` is called prior to calling the `fit` method in `StackingRegressor` and `StackingCVRegressor`. ([#315](https://github.com/rasbt/mlxtend/issues/315))
 - The `plot_decision_regions` function now automatically determines the optimal setting based on the feature dimensions and supports anti-aliasing. The old `res`  parameter has been deprecated. ([#309](https://github.com/rasbt/mlxtend/pull/309) by [Guillaume Poirier-Morency](https://github.com/arteymix))
 - Apriori code is faster due to optimization in `onehot transformation` and the amount of candidates generated by the `apriori` algorithm. ([#327](https://github.com/rasbt/mlxtend/pull/327) by [Jakub Smid](https://github.com/jaksmid))
+- `onehot transform` can be now called with `sparse` argument resulting in the sparse representation of the `onehot` matrix
```
rasbt: You can append 

([#328](https://github.com/rasbt/mlxtend/pull/328) by [Jakub Smid](https://github.com/jaksmid))

jaksmid:
Thanks for the code review @rasbt .

Incorporated your suggestions. What about the documentation in the notebooks though?

Should I address that as well?

rasbt:
Good point. If you would like to add an example to the Jupyter Notebook, that would be great (but not 100% necessary). To update the API documentation in the notebooks, you need to go to 

mlxtend/docs and execute 

```python make_api.py```

in this directory. It will then update/generate the markdown documentation files. After that's completed, you can open the jupyter notebook and just rerun the last section of the notebook.

As a note, these markdown files are not added to the version control in GitHub because of redundancy (i.e., the API doc markdown gets included in the notebook at the bottom), and the markdown generated from the notebook for MKdocs has the same content as the notebook (so, for space reasons, only the notebooks are currently in version control on GitHub).

**Comment by jaksmid** [comment]
Will do. 

Was also thinking about changing the `dtype` to `bool` in the `onehot.py`. This can further reduce the memory footprint

**Comment by jaksmid** [comment]
Hello @rasbt ,

updated the doc, added a `sparse` example to the `apriori` notebook  and changed the type of `onehot transforms` to bool. I manually verified that the `apriori` is still working as expected. Tests also reported no problem.

Looking forward to the feedback.

**Comment by rasbt** [comment]
Oh, I somehow assumed we were talking about the OneHotTransaction class regarding the Jupyter notebook documentation.

But yeah, modifying the apriori documentation is even better :). 

However, it would maybe be nice to update the OneHotTransaction documentation as well (and then add a note about why it's using bool by default, and maybe showing an `astype` examples to get the classic binary integer representation -- sorry if that's too much asking, I would also be happy to do that :).

Overall, this PR is great, and it was a good idea to use sparse representations for this!

**Comment by rasbt** [comment]
Just a little suggestion, but can you modify the Changelog message a little bit to illustrate that the apriori func also "improves"?

E.g., changing the original

`onehot transform` is now 
more memory efficient as it uses boolean instead of ints. Furthermore, `onehot transform` can be now called with `sparse` argument resulting in the sparse representation of the `onehot` matrix. ([#328](https://github.com/rasbt/mlxtend/pull/328) by [Jakub Smid](https://github.com/jaksmid))

to

The `OnehotTransactions` class (which is typically often in combination with the `apriori` function for associaton rule mining) is now more memory efficient as it uses boolan arrays instead of integer arrays. In addition, the `OnehotTransactions` class can be now be provided with `sparse` argument to generate sparse representations of the `onehot` matrix to further improve memory efficiency. ([#328](https://github.com/rasbt/mlxtend/pull/328) by [Jakub Smid](https://github.com/jaksmid))


Below is a rawtext version for convenience :)

```
The `OnehotTransactions` class (which is typically often in combination with the `apriori` function for associaton rule mining) is now more memory efficient as it uses boolan arrays instead of integer arrays. In addition, the `OnehotTransactions` class can be now be provided with `sparse` argument to generate sparse representations of the `onehot` matrix to further improve memory efficiency. ([#328](https://github.com/rasbt/mlxtend/pull/328) by [Jakub Smid](https://github.com/jaksmid))
```

jaksmid:
Not a problem at all.

I tried to find all places that needs/deserves an update, but as I am not that familiar with the whole codebase, it may well happen that I unintentionally overlook something.

Really appreciate your suggestions.

rasbt:
Looks great, thanks a lot! Will merge it now, thanks for the PR!

PR #342, opened 2018-03-13
Adds style related dictionaries to plot_decision_regions

### LINKED ISSUE #340: Add markersize parameter to plot_decision_regions
I recently needed to change the scatter plot marker size in `plot_decision_regions` to make the markers easier to see in a presentation. A relatively easy fix was to use the matplotlib style context manager to temporarily change the plotting style. For example, 

```python 
with plt.style.context({'lines.markersize': 50}):
        plot_decision_regions(X, y, clf, colors='C0,C1,C2', markers='ooo' , ax=ax)
```

However, given that `plot_decision_regions` already has a `markers` and `colors` parameter for adjusting the marker style, I thought it would be natural to have a `markersize` parameter to more easily adjust the scatter plot marker size. 

@rasbt does adding a `markersize` parameter to `plot_decision_regions` sound reasonable? If so, I can submit a PR.

jrbourbeau:
### Description

This PR adds parameters to `plot_decision_regions` that can be used to customize the plotting style used (e.g. size of scatter plot markers, alpha value of decision regions, etc.)

Fixes #340

### Pull Request requirements


jrbourbeau:
Additional TODO:

- [ ] Make user can't override `markers` and `colors` parameters in `plot_decision_regions` with `scatter_kwargs`, etc.

mlxtend/plotting/decision_regions.py
```diff
@@ -270,13 +281,59 @@ def plot_decision_regions(X, y, clf,
...
+            pass
+        else:
+            handles, labels = ax.get_legend_handles_labels()
+            ax.legend(handles, labels,
+                      framealpha=0.3, scatterpoints=1, loc=legend)
 
     return ax
+
+
+def format_plotting_kwargs(default_kwargs=None, user_kwargs=None,
```
rasbt: That's a nice function dealing with the issues of protected keys! One little thing that comes to mind is that the way I wrote the API parser for the documentation, this would show up in the API docs. You could make this function private by prefixing it with an underscore (`_format_plotting_kwargs`), but I think this function can come in very handy in other functions (esp. plotting functions) as well in future. So, it would be nice if you could move it to the `mlxtend.utils` :)

rasbt: I don't have a better name at hand, but under a more general name like "format_kwarg_dictionaries" or so?

jrbourbeau: Sounds good to me! I have a couple of quick questions:

1) Should this function go into its own module in `utils`, maybe `plotting.py`, or into one of the existing modules like `checking.py`?

2) Should I also add the leading underscore to make this function private in `utils` (i.e. `_format_kwarg_dictionaries` vs. `format_kwarg_dictionaries`)? I'd imagine this function will only be used for internal stuff and not directly by users, so the leading underscore would make sense. 

Thanks!

rasbt: Good points! I think we could put it into checking, because it's somewhat related to checking kwargs. I wouldn't suggest making it private, because it's generally useful (how I think about private things is that these are more like helper functions that are not useful outside the scope of a particular function and more like a side effect of refactoring, whereas the public ones are more broadly useful).

jrbourbeau:
I can also add a `plot_decision_regions` user guide example to illustrate how to use the style kwargs

rasbt:
> I can also add a plot_decision_regions user guide example to illustrate how to use the style kwargs

Thanks, that would be super nice :)

jrbourbeau:
Unless you have any other comments, I think this PR should be good to go!

rasbt:
Based on the code, this looks great, thanks so much! I will take a look at the notebook as well and give it a try, but it looks perfect so far!

rasbt:
Looks and works perfectly, thanks for the great contribution!

Issue #344, opened 2018-03-14
Change itemsets generated via `apriori`  from list to sets


rasbt:
I think it would be good to change the itemset representation generated via `apriori` from a list representation to a set representation, i.e.,

![screen shot 2018-03-14 at 9 23 11 am](https://user-images.githubusercontent.com/5618407/37405580-6678ee24-276b-11e8-934c-75fa24e9f182.png)

   |
   |
   v

![screen shot 2018-03-14 at 9 31 36 am](https://user-images.githubusercontent.com/5618407/37405582-67d9a6b4-276b-11e8-8a20-bde91b5475ce.png)

For reasons discussed in #343

WLaney:
If people are still on board with this enhancement I can take a stab at implementing it

rasbt:
Thanks. I think it may be an improvement overall :) . (But could be that I am missing something.)

WLaney:
Yeah I agree. I'll get to work implementing this

rasbt:
addressed in #393

PR #393, opened 2018-06-11
Apriori use sets

WLaney:
### Description
Changes `apriori` to return itemsets as sets instead of lists.

Disscused in [issue 344](https://github.com/rasbt/mlxtend/issues/344)

### Pull Request Checklist

### DIFF
--- a/mlxtend/frequent_patterns/tests/test_apriori.py
+++ b/mlxtend/frequent_patterns/tests/test_apriori.py
@@ -57,8 +57,8 @@ def test_max_len():
 def test_itemsets_type():
     res_colindice = apriori(df, use_colnames=False)  # This is defualt behavior
     for i in res_colindice['itemsets']:
-        assert isinstance(i, set) == True
-        
+        assert isinstance(i, set) is True
+
     res_colnames = apriori(df, use_colnames=True)  # This is defualt behavior
     for i in res_colnames['itemsets']:
-        assert isinstance(i, set) == True
+        assert isinstance(i, set) is True
--- a/mlxtend/frequent_patterns/tests/test_apriori.py
+++ b/mlxtend/frequent_patterns/tests/test_apriori.py
@@ -52,3 +52,13 @@ def test_max_len():
 
     res_df2 = apriori(df, max_len=2)
     assert len(res_df2.iloc[-1, -1]) == 2
+
+
+def test_itemsets_type():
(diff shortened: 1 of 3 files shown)


rasbt:
Thanks for the PR! 

I think for efficiency reasons, to avoid yet another for-loop, it would be better to to the transformation to sets some place earlier in the code. For example, replacing the line `itemsets = pd.Series([set(i) for i in itemset_dict[k]]` by

```
itemsets = pd.Series([set(i) for i in itemset_dict[k]]
```

and then the line for the `use_colnames` loop as well. I.e., `res_df['itemsets'] = res_df['itemsets'].apply(lambda x: {mapping[i]
                                                      for i in x})`.

Would be great if you could also add a short unit test checking that the types of the itemsets is indeed "set" for `use_colnames=True` & `False`

Thanks!

WLaney:
Thank you for the advice!

I tried to do the transformation earlier in the code on my first go, but I couldn't quite figure out where to do it without messing up the panda data structure. I was able to get it this time with your help though!

For the unit test I'm just looping through the returned itemsets and checking the type of each element. I've never had to make unit tests before, but I that test seems to checks what we want. If you think it would be better to do something else though just let me know.

mlxtend/frequent_patterns/tests/test_apriori.py
```diff
@@ -52,3 +52,13 @@ def test_max_len():
 
     res_df2 = apriori(df, max_len=2)
     assert len(res_df2.iloc[-1, -1]) == 2
+
+
+def test_itemsets_type():
+    res_colindice = apriori(df, use_colnames=False)  # This is defualt behavior
```
rasbt: Could you change "defualt" to "default" ? :)

mlxtend/frequent_patterns/tests/test_apriori.py
```diff
@@ -52,3 +52,13 @@ def test_max_len():
...
+    for i in res_colindice['itemsets']:
+        assert isinstance(i, set) is True
+
+    res_colnames = apriori(df, use_colnames=True)  # This is defualt behavior
```
rasbt: I think this shouldn't be default behavior then, because the line above with "use_colnames=False" is indicated with default behavior

rasbt:
That looks good, thanks! Could you maybe just do the little comment-fix that I mentioned above? Besides that,  this looks good!

rasbt:
That's great, thanks for the fix & PR

PR #441, opened 2018-09-23
Adds sample_weight option to estimators fit method

### LINKED ISSUE #438: Support for `sample_weight` in the `fit` method
Is there any plan or demand for supporting `sample_weight` option in the `fit` method of estimators?  

Since a number of model classes in scikit-learn support it, I suppose it would be useful for estimators such as `StackingRegressor` (provided that all regressors accept it).

I saw a related PR #255 about the support for `fit_params` options.  Once this is merged, `sample_weight` will be implementable by passing same weights to all regressors and meta-regressor.  Even so, API like `fit(X, y [, sample_weight])` can be handy.

kota7:
### Description

Adds `sample_weight` option to the `fit` method of estimators.
Aims to cover the followings:

- [x] StackingClassifier
- [x] StackingCVClassifier
- [x] EnsembleVoteClassifier
- [x] StackingRegressor
- [x] StackingCVRegressor

Fixes #438 

### Pull Request Checklist

### DIFF
--- a/mlxtend/regressor/stacking_regression.py
+++ b/mlxtend/regressor/stacking_regression.py
@@ -101,7 +101,7 @@ class StackingRegressor(BaseEstimator, RegressorMixin, TransformerMixin):
         self.store_train_meta_features = store_train_meta_features
         self.refit = refit
 
-    def fit(self, X, y):
+    def fit(self, X, y, sample_weight=None):
         """Learn weight coefficients from training data for each regressor.
 
         Parameters
@@ -111,6 +111,8 @@ class StackingRegressor(BaseEstimator, RegressorMixin, TransformerMixin):
             n_features is the number of features.
         y : array-like, shape = [n_samples] or [n_samples, n_targets]
             Target values.
+        sample_weight : array-like, shape = [n_samples], optional
+            Sample weights.
 
         Returns
         -------
@@ -141,16 +143,18 @@ class StackingRegressor(BaseEstimator, RegressorMixin, TransformerMixin):
             if self.verbose > 1:
                 print(_name_estimators((regr,))[0][1])
 
-            regr.fit(X, y)
+            regr.fit(X, y, sample_weight=sample_weight)
 
         meta_features = self.predict_meta_features(X)
 
         if not self.use_features_in_secondary:
-            self.meta_regr_.fit(meta_features, y)
+            # meta model uses the prediction outcomes only
+            pass
         elif sparse.issparse(X):
-            self.meta_regr_.fit(sparse.hstack((X, meta_features)), y)
+            meta_features = sparse.hstack((X, meta_features))
         else:
-            self.meta_regr_.fit(np.hstack((X, meta_features)), y)
+            meta_features = np.hstack((X, meta_features))
+        self.meta_regr_.fit(meta_features, y, sample_weight=sample_weight)
 
         # save meta-features for training data
         if self.store_train_meta_features:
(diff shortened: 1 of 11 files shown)


mlxtend/regressor/stacking_regression.py
```diff
@@ -111,6 +111,8 @@ def fit(self, X, y):
             n_features is the number of features.
         y : array-like, shape = [n_samples] or [n_samples, n_targets]
             Target values.
+        sample_weight : array-like, shape = [n_samples], optional
+            Sample weights.
```
rasbt: Could you additionally specify that these are used by both the level-1 and meta-regressors?

kota7: Sure.  And the meta regressor too.

rasbt: Yeah, maybe as doc string sth like

Sample weights passed as `sample_weights` to each regressor in the `regressors` list as well as the `meta_regressor`.

The rest of the PR looks great btw! Thanks!

kota7:
@rasbt This thread may not be the best place to talk about this, but when I update the test script `test_stacking_cv_regressor.py`, I found that generating another random vector make existing tests fail.

For example, below this part:
https://github.com/rasbt/mlxtend/blob/030c1b703ffd52aef7d6b4a6e003451edd02f4fb/mlxtend/regressor/tests/test_stacking_cv_regression.py#L22-L30

if you add another line:
```
w = np.random.random(40)
```
then tests start to fail.  This is perhaps because cross validation generators share random seed with numpy.
To see this, the below code generates two different results, due to the call for `np.random.random` after setting seed.

```
import numpy as np
from sklearn.model_selection import KFold

np.random.seed(1)
cv = KFold(2, shuffle=True)
print(list(cv.split([1,2,3,4,5,6])))

np.random.seed(1)
np.random.random(10)
cv = KFold(2, shuffle=True)
print(list(cv.split([1,2,3,4,5,6])))
```

As a side effect, when I some test raises exception, then the random seed changes, which makes other tests to fail.  Due to this, error report becomes confusing since tests that have no problem fail.

A possible workaround is to explicitly define cross validation object with specific random state at the top of script, and keep using it in tests.  Would you want me to do this, or do you have any thoughts?

rasbt:
Thanks for bringing this up. This is actually a bit of a messy unit test design. The random seed should either be set for each function individually or it should be using the `random_state` via scikit-learn, which is probably the better solution.

I think the reason why the StackingCVRegressor currently doesn't have a `random_state` parameter is because it uses the `sklearn.model_selection.check_cv` method, which doesn't support that either. 

So, one option would be resetting the random seed in each unit test seems to be the best option for now, but that might be a lot of work, unfortunately, because the unit tests might produce different results then like you mentioned.

Your suggestion

> A possible workaround is to explicitly define cross validation object with specific random state at the top of script, and keep using it in tests.

is probably better because in case we change the the StackingCVRegressor in future to have its own random_state, a fixed Kfold object wouldn't require changing all the unit tests again

kota7:
@rasbt I think it is ready for review. Thanks.

rasbt:
This looks really nice overall. I am wondering though...,  don't the sklearn estimators accept `sample_weight=None`? This could simplify a lot of the if/else statements that are currently in there:

            regr.fit(X, y)		             if sample_weight is None:
                 regr.fit(X, y)
             else:
                 regr.fit(X, y, sample_weight=sample_weight)

kota7:
As far as I know, `Lasso`, `MLPClassifier`, `KNerighborsClassifier` do not support it, hence raises exception if you give `sample_weight` to them.  This piece of unittest catches that:

https://github.com/rasbt/mlxtend/blob/b46079a21209220f92ee274f4498e120e9a1b762/mlxtend/regressor/tests/test_stacking_regression.py#L132-L146

That is, if you code just `regr.fit(X, y, sample_weight=sample_weight)`, `sample_weight=None` is passed to `Lasso`, which raises error.

rasbt:
Good point! Otherwise the PR seems fine and I'd be happy to merge. Or do you have any additions in mind?

rasbt:
Awesome! Thanks for this PR, really appreciate it!

PR #446, opened 2018-09-27
minor correction and extension of doc on lift score calculation

JanLauGe:
### Description

I've noticed a minor error in the documentation for lift score calculation. The union in $support(true labels∪prediction)/N$ should actually be an intersect. This is correctly implemented in the code but also incorrect on the referenced Wikipedia article. I have changed it here and also modified a few other points in the doc:

- modified confusion matrix based lift formula. The numerator is now identical with model precision. Result remains identical.
- modified support based lift formula. This previously included N. However, with the way that support is implemented in the code, division by N is already part of the support calculation and should therefore not be listed separately in the equation.
- updated results so that number of decimal places shown matches up between the different examples.

### Pull Request Checklist

### DIFF
--- a/docs/sources/user_guide/evaluate/lift_score.ipynb
+++ b/docs/sources/user_guide/evaluate/lift_score.ipynb
@@ -268,21 +268,21 @@
  "metadata": {
   "anaconda-cloud": {},
   "kernelspec": {
-   "display_name": "Python 2",
+   "display_name": "Python 3",
    "language": "python",
-   "name": "python2"
+   "name": "python3"
   },
   "language_info": {
    "codemirror_mode": {
     "name": "ipython",
-    "version": 2
+    "version": 3
    },
    "file_extension": ".py",
    "mimetype": "text/x-python",
    "name": "python",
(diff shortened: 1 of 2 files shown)


rasbt:
Thanks a lot, I really appreciate it that you are looking into that. 
Regarding the unit tests, I think the failure may be related to scikit-learn's 0.2 version that was released yesterday. I.e., the "Latest" ones prob already download 0.2.0 from conda.  When I see it correctly, your fixes only apply to the documentation, so that shouldn't affect the unit tests

screen shot 2018-09-27 at 12 29 58 pm

JanLauGe:
Thank YOU for all the work on the package, and no worries, it was a good refresher digging into the details of this. Yes, changes are to the documentation only (and some changes resulting from re-running the notebook cell with a different python version x) )

rasbt:
The initial implementation was based on Thanks @JanLauGe I had on my to-do list to double check with some official resource but couldn't find anything good and went back to my class notes. I believe this should be correct now -- or at least it reflects what's actually implemented in the code :). Thanks a lot!

PR #479, opened 2018-12-07
Update USER_GUIDE_INDEX.md broken links

CatcherGG:
### Description

User guide index contains links to md files. Those files are non-existent.
Instead it should direct the correct ipynb files.


rasbt:
Thanks! I think there's a little misunderstanding :). This is actually an internal document. To make the docs, I have a script that converts the notebooks to markdown, and the USER_GUIDE_INDEX.md file is then used by mkdocs to make the HTML. That's why it has links to .md files. I.e,. the website automatically rendered from this would be this:

http://rasbt.github.io/mlxtend/USER_GUIDE_INDEX/

Issue #485, opened 2018-12-21
Confusion matrix with class names


Juanlu001:
I came here looking for a quick way to plot a confusion matrix with class labels, but found that there's not such a way. I think it would be a great addition, at least to the documentation. For the moment, I'll be using scikit-plot.

rasbt:
Good point, I can add it to the documentation. I could add it as function parameters, but it's maybe overkill. The modification would basically be 

```python
fig, ax = plot_confusion_matrix(conf_matrix)
ax.set_xticklabels(['', 'Positive Class (0)', 'Negative Class (1)'])
ax.set_yticklabels(['', 'Positive Class (0)', 'Negative Class (1)']);
```

rasbt:
I see. You mean if you don't have integer class labels like

```python
y_target =    [1, 1, 1, 0, 0, 2, 0, 3]
y_predicted = [1, 0, 1, 0, 0, 2, 1, 3]
```
but some string variables in unknown order? In that case it will be ordered alphabetically.


import matplotlib.pyplot as plt
from mlxtend.evaluate import confusion_matrix
import numpy as np

```python
y_target =    ['b', 'b', 'b', 'a', 'a', 'c', 'a', 'd']
y_predicted = ['b', 'a', 'a', 'a', 'a', 'c', 'a', 'd']

cm = confusion_matrix(y_target=y_target, 
                      y_predicted=y_predicted,)
cm
```

```array([[3, 0, 0, 0],
       [2, 1, 0, 0],
       [0, 0, 1, 0],
       [0, 0, 0, 1]])
```

and to plot it:

```python
from mlxtend.plotting import plot_confusion_matrix

fig, ax = plot_confusion_matrix(cm)

ax.set_xticklabels([''] + list(np.unique(y_target)))
ax.set_yticklabels([''] + list(np.unique(y_target)))
plt.show()
```

![unknown](https://user-images.githubusercontent.com/5618407/50352822-026daf80-050c-11e9-92ff-f211e3884e54.png)

rasbt:
Has been addressed now via #487

Juanlu001:
Awesome, thanks!

PR #487, opened 2019-01-01
Add class names param to plot_confusion_matrix

sandpiturtle:
### Pull Request Checklist

- [ ] Ran `nosetests ./mlxtend -sv` and make sure that all unit tests pass (for small modifications, it might be sufficient to only run the specific test file, e.g., `nosetests ./mlxtend/classifier/tests/test_stacking_cv_classifier.py -sv`)


rasbt:
Thanks for the PR. This is a nice addition according to the discussion in #485 

Since there are no unit tests for plotting visualizations, could you please add an example in the documentation (at the bottom of https://github.com/rasbt/mlxtend/blob/master/docs/sources/user_guide/plotting/plot_confusion_matrix.ipynb ; these jupyter notebooks are used to render the HTML documentation upon a new version release)

Also, it would be nice if you could make an entry in the Changelog.

rasbt:
I can totally understand your motivation behind that, and I agree that it could add some more convenience. However, I sometimes think that adding more bells and whistles (more options) can also complicate the use -- when people use their own random seeds, they might wonder where/how exactly the global random seed in the stacking class is applied exactly, and what it overrides and what not. So, I suggest sticking with having users define the random seed on the algorithm level. I think, this could then also avoid issues when people using some algorithms that don't have a `random_state` parameter (e.g., if the algorithm doesn't require one -- I guess KNN and decision trees would be typical examples -- or the `random_state` attribute is named `random_seed` for some reason)

PR #500, opened 2019-02-10
Applying to iloc of pandas data frame

tetrar124:
### Description
Bug fix for selecting columns and add new features for the name of the columns of pandas data frame.

### Pull Request Checklist

### DIFF
--- a/mlxtend/feature_selection/column_selector.py
+++ b/mlxtend/feature_selection/column_selector.py
@@ -83,7 +83,7 @@ class ColumnSelector(BaseEstimator):
             types = {type(i) for i in self.cols}
             if len(types) > 1:
                 raise ValueError('Elements in `cols` should be all of the same data type.')
-            if isinstance(self.cols[0], int):
+            elif isinstance(self.cols[0], int):
                 t = X.iloc[:, self.cols].values
             elif isinstance(self.cols[0], str):
                 t = X.loc[:, self.cols].values
--- a/mlxtend/feature_selection/column_selector.py
+++ b/mlxtend/feature_selection/column_selector.py
@@ -83,7 +83,7 @@ class ColumnSelector(BaseEstimator):
             types = {type(i) for i in self.cols}
             if len(types) > 1:
                 raise ValueError('Elements in `cols` should be all of the same data type.')
-            elif isinstance(self.cols[0], int):
+            if isinstance(self.cols[0], int):
                 t = X.iloc[:, self.cols].values
             elif isinstance(self.cols[0], str):
                 t = X.loc[:, self.cols].values
--- a/mlxtend/feature_selection/column_selector.py
+++ b/mlxtend/feature_selection/column_selector.py
@@ -77,11 +77,18 @@ class ColumnSelector(BaseEstimator):
         """
 
         # We use the loc accessor if the input is a pandas dataframe
-        if hasattr(X, 'loc'):
-            try:
+        if hasattr(X, 'loc') or hasattr(X, 'iloc'):
+            if type(self.cols) == tuple:
+               self.cols = list(self.cols)
+            types = {type(i) for i in self.cols}
+            if len(types) > 1:
+                raise ValueError('Elements in `cols` should be all of the same data type.')
+            if isinstance(self.cols[0], int):
                 t = X.iloc[:, self.cols].values
-            except TypeError:
+            elif isinstance(self.cols[0], str):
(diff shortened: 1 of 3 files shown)


rasbt:
Thanks for the PR. However, instead doing a try-except, I think the following could solve the problem better? (Haven't tested it, but what do you think? Does that cover your use case issue?)

```python
        if hasattr(X, 'loc') or hasattr(X, 'iloc'):
            types = {type(i) for i in self.cols}
            if len(types) > 1:
                raise ValueError('Elements in `cols` should be all of the same data type.')
            if isinstance(cols[0], int):
                t = X.iloc[:, self.cols].values
            elif isinstance(cols[0], str):
                t = X.loc[:, self.cols].values
           else:
               raise ValueError('Elements in `cols` should be either `int` or `str`.')
```

tetrar124:
I checked your code on my environment, your code works well (I changed only cols[0] to self.cols[0]). And I added 2 lines in this code for using tuple because of using a tuple in the user's guide.
http://rasbt.github.io/mlxtend/user_guide/feature_selection/ColumnSelector/
```
        if hasattr(X, 'loc') or hasattr(X, 'iloc'):
            if type(self.cols) == tuple:
               self.cols = list(self.cols)
            types = {type(i) for i in self.cols}
            if len(types) > 1:
                raise ValueError('Elements in `cols` should be all of the same data type.')
            if isinstance(self.cols[0], int):
                t = X.iloc[:, self.cols].values
            elif isinstance(self.cols[0], str):
                t = X.loc[:, self.cols].values
            else:
               raise ValueError('Elements in `cols` should be either `int` or `str`.')
```
Thank you for your cooperation.

rasbt:
Thanks for working on this PR! Just as a tip to make the debugging smoother/faster (since the CI is very slow), you can also test your code locally by running

    nosetests mlxtend/feature_selection

to see if there are any issues with the code.

mlxtend/feature_selection/column_selector.py
```diff
@@ -83,7 +83,7 @@ def transform(self, X, y=None):
             types = {type(i) for i in self.cols}
             if len(types) > 1:
                 raise ValueError('Elements in `cols` should be all of the same data type.')
-            if isinstance(self.cols[0], int):
+            elif isinstance(self.cols[0], int):
```
rasbt: I don't think this should be an elif here because we want to check for both things

- first that the column names have the same type
- if the above is true, the second check will check whether these are integer column types.

mlxtend/feature_selection/column_selector.py
```diff
@@ -75,12 +76,22 @@ def transform(self, X, y=None):
...
+        if hasattr(X, 'loc') or hasattr(X, 'iloc'):
+            if type(self.cols) == tuple:
+                self.cols = list(self.cols)
+            types = {type(i) for i in self.cols}
+            if len(types) > 1:
+                raise ValueError('Elements in `cols` should be all of the same data type.')
+            elif isinstance(self.cols[0], int):
+                t = X.iloc[:, self.cols].values
+            elif isinstance(self.cols[0], str):
+                t = X.loc[:, self.cols].values
```
rasbt: I don't think this should be an elif here (can you please change it back) because we want to check for both things

* first that the column names have the same type
* if the above is true, the second check will check whether these are integer column types.

tetrar124:
Sorry, I return elif to if. The reason for using elif is only for the coverage. I checked new lines with nose and flask8, and there are no errors other than 'line too long' from the new comment lines for ValueError.

rasbt:
Ok thanks :)! Btw don't worry about minor changes regarding the coverage too much. Can you please make the line < 79 chars though (splitting it over two lines) -- we decided to strictly enforce PEP8 at some point so that the library won't become too messy over time by making exceptions.

rasbt:
Could you also add an additional unit test for using a DataFrame with integer column names to make sure it works correctly? (I think we don't have that one yet) :)

tetrar124:
I made the data frame with columns whose names were only int values, and I checked the type of these columns names showed int.
When the name of a column was int values, there was no bug but this code has worked only as the mode of iloc.

rasbt:
Thanks for testing. We should add a unit test though so that we can ensure that the behavior doesn't break in future. 

I just added a unit test and changelog entry to this PR. Let me know if you have any further comments. Otherwise, I think it's good to merge if/when the CI tests pass. Thanks for the PR!

PR #523, opened 2019-04-29
Add random_state parameter to stacking cv estimators

qiagu:
### Description
The idea just came to my mind today. Since the latest `check_cv` from scikit-learn supports `random_state`, stacking CV estimators can have the parameter now. 

### Pull Request Checklist


rasbt:
Great point. I think this is a relatively new feature and I didn't know it would work. While this is certainly great, one little request though, 

>     random_state : int, RandomState instance or None, optional (default: 0)
         Constrols the randomness of the cv splitter. Used when `cv` is
         integer and `shuffle=True`. New in v0.16.0.

Can we change that to `None` as default setting (while I personally prefer setting random seeds explicitly, I think this is a scikit-learn default to use None.)

mlxtend/regressor/stacking_cv_regression.py
```diff
@@ -115,7 +115,7 @@ class StackingCVRegressor(_BaseXComposition, RegressorMixin, TransformerMixin):
     def __init__(self, regressors, meta_regressor, cv=5,
                  shuffle=True, random_state=0, verbose=0,
                  refit=True, use_features_in_secondary=False,
-                 store_train_meta_features=False, n_jobs=1,
+                 store_train_meta_features=False, n_jobs=None,
```
rasbt: I actually meant the `random_state` to be `random_state=None`, but good that you caught the `n_jobs=None` thing as well (which is another sklearn convention)

qiagu:
The `random_state` default was given by not `None` because the `shuffle` was `True` and also it may save typing in many cases. I'm fine with `None` and don't have strong inclinations. I'll follow you decision. @rasbt

rasbt:
I see. I think `shuffle=True` would still work with `random_state=None` -- as far as I know, `random_state=None` will just use the default random seed. I don't have a strong preference, but for consistency with sklearn, maybe `random_state=None` is the slightly better choice.

qiagu:
I agree. Updated.

rasbt:
That's great, happy to merge this. Thanks a lot!

Issue #543, opened 2019-05-27
PCA Correlation circle


Gabriel-Azevedo-Ferreira:
I would like to add a correlation circle graph for PCA, like the one in R's FactormineR.

![nba](https://user-images.githubusercontent.com/19335547/58408689-e1078580-80a0-11e9-950b-18f992265c25.png)

rasbt:
This would be a great addition! It could be added to the mlxtend.plotting submodule.

Gabriel-Azevedo-Ferreira:
Hello @rasbt ,
Thanks! I implemented it and made a pull request.
The pull request failed. Looking at the reasons I found the following:
   - Travis: `ERROR: Failure: ModuleNotFoundError (No module named 'adjustText')`
           - This package was not strictly necessary, it basically rearranges the text in the figure in order to avoid conflicts and overlaps. I can remove the use of it in a next pull request, but, if there is a way to keep it, I think it would be good. What do you suggest me to do? How can I make Travis "understand" that it should install "adjustText" for the test?
   - appveyor: It seems that the error was due to other tests, not related to my submission. Does this make sense? It seems the failed tests are in "mlxtend/regressor/tests" and  "mlxtend/classifier/tests". It seems that the tests in  "mlxtend/plotting/tests" were not ran.

rasbt:
Thanks! 

> In order to keep it, I added a line on ci/.travis_install.sh: pip install adjustText.

this would not be enough, I think, because then later the conda-forge build will also have problems. In that case, we would also have to add it to the setup.py requirements then.

But regarding the adjustText package, it would be good to avoid additional installation requirements and dependencies (since we already have so many) if possible. In that case, it looks like AdjustText is just a 1-file library when I see that correctly? (https://github.com/Phlya/adjustText). We could simply add it to the mlxtend.externals subpackage since it's licensed under MIT license, which is compatible with mlxtend. I.e., similar to the `six` library (https://github.com/rasbt/mlxtend/blob/master/mlxtend/externals/six.py) with the license file on top.

Could you change that?

rasbt:
Added via #544

BrandonKMLee:
@rasbt thanks for including this

BrandonKMLee:
@rasbt is it possible to put a scatter plot on top of the correlation circle for interpretability purposes?

PR #544, opened 2019-05-28
Adding PCA correlation circle graph

Gabriel-Azevedo-Ferreira:
### Description

Adding a function to enable the plotting of PCA correlation circles, similarly to what is done by factomineR in R.

Implements the feature proposed in #543 (https://github.com/rasbt/mlxtend/issues/543)

### Pull Request Checklist

- [DONE] Added a note about the modification or contribution to the `./docs/sources/CHANGELOG.md` file (if applicable)
- [DONE ] Added appropriate unit test functions in the `./mlxtend/*/tests` directories (if applicable)
- [DONE] Modify documentation in the corresponding Jupyter Notebook under `mlxtend/docs/sources/` (if applicable)
- [DONE] Ran `nosetests ./mlxtend -sv` and make sure that all unit tests pass (for small modifications, it might be sufficient to only run the specific test file, e.g., `nosetests ./mlxtend/classifier/tests/test_stacking_cv_classifier.py -sv`)
- [DONE] Checked for style issues by running `flake8 ./mlxtend`


docs/sources/CHANGELOG.md
```diff
@@ -19,7 +19,7 @@ The CHANGELOG for the current development version is available at
 - Added an enhancement to the existing `iris_data()` such that both the UCI Repository version of the Iris dataset as well as the corrected, original
   version of the dataset can be loaded, which has a slight difference in two data points (consistent with Fisher's paper; this is also the same as in R). (via [#539](https://github.com/rasbt/mlxtend/pull/532) via [janismdhanbad](https://github.com/janismdhanbad))
 - Add optional `groups` parameter to `SequentialFeatureSelector` and `ExhaustiveFeatureSelector` `fit()` methods for forwarding to sklearn CV ([#537](https://github.com/rasbt/mlxtend/pull/537) via [arc12](https://github.com/qiaguhttps://github.com/arc12))
-
+- Adding plotting PCA correlation graph
```
rasbt: I suggest extending this

- Added a new `plot_pca_correlation_graph` function to the `mlxtend.plotting` submodule for plotting a PCA correlation graph. ([#544](https://github.com/rasbt/mlxtend/pull/544) via [Gabriel-Azevedo-Ferreira](https://github.com/qiaguhttps://github.com/Gabriel-Azevedo-Ferreira))

ci/.travis_install.sh
```diff
@@ -51,6 +51,7 @@ if [ "${COVERAGE}" = "true" ]; then
 fi
 
 pip install nose-exclude
+pip install adjustText
```
rasbt: As discussed, we could add the `adjustText.py` file to `mlxtend.externals`

mlxtend/plotting/__init__.py
```diff
@@ -17,6 +17,7 @@
 from .checkerboard import checkerboard_plot
 from .ecdf import ecdf
 from .scatterplotmatrix import scatterplotmatrix
+from .corr_circle_pca import plot_pca_correlation_graph
```
rasbt: The file could be named `pca_correlation_graph.py` instead of `corr_circle_pca.py`, which could make it slightly easier to find & relate to the imported function

mlxtend/plotting/corr_circle_pca.py
```diff
@@ -0,0 +1,97 @@
...
+# File Author: Gabriel Azevedo Ferreira <az.fe.gabriel@gmail.com>
+
+from sklearn.decomposition import PCA
+import numpy as np
+import matplotlib.pyplot as plt
+import pandas as pd
+from adjustText import adjust_text
+
+
+def create_correlation_table(A, B, names_cols_A, names_cols_B):
```
rasbt: Could you use the NumPy docstring convention for docstrings? Otherwise, the documentation will not be created properly (the documentation building tools rely on that particular docstring format). See the other docstrings for examples how they should look like.

mlxtend/plotting/corr_circle_pca.py
```diff
@@ -0,0 +1,97 @@
+# mlxtend Machine Learning Library Extensions
+#
+# A function for plotting a PCA correlation circle
+# File Author: Gabriel Azevedo Ferreira <az.fe.gabriel@gmail.com>
+
+from sklearn.decomposition import PCA
```
rasbt: there's also a PCA in mlxtend that I added a long time ago:

    from mlxtend.feature_extraction import PrincipalComponentAnalysis

They should provide equivalent results (both rely on SVD in scipy).

mlxtend/plotting/corr_circle_pca.py
```diff
@@ -0,0 +1,97 @@
...
+    variables_names : array like
+        Name of the columns (the variables) of X
+    dimensions: tuple with two elements.
+        dimensions to be plot (x,y)
+    X_pca : optional. if not provided, compute PCA independently
+    figure_axis_size :
+         size of the final frame. The figure created is a square with length
+         and width equal to figure_axis_size.
+
+    returns:
```
rasbt: Also here, the `returns` should be formatted like the other docstrings so that the documentation will be formatted ok

mlxtend/plotting/corr_circle_pca.py
Gabriel-Azevedo-Ferreira: I have just tried to replace it and the results seemed to be different. Maybe I need to tweak a little the output or provide a slightly different input. The code for each of the pictures is the same one I put in the docs in the Jupyter notebook (normalise the columns of iris and apply PCA). 

If it is ok, I will not change this for now and later try to figure out which one of the two is correct.

mlxtend  - PrincipalComponentAnalysis
image

sklearn PCA

rasbt: It's a long time ago, but as far as I remember, the reason why I implemented my own PCA is that back then, scikit-learn didn't store the eigenvalues (or sth like that). Just checking, the mlxtend and scikit-learn impl. indeed produce the same results:

```python
from mlxtend.data import iris_data
from mlxtend.preprocessing import standardize
from mlxtend.feature_extraction import PrincipalComponentAnalysis
import numpy as np
import matplotlib.pyplot as plt

X, y = iris_data()
X = standardize(X)

pca = PrincipalComponentAnalysis(n_components=None)
pca.fit(X)
X_pca = pca.transform(X)

tot = sum(pca.e_vals_)
var_exp = [(i / tot)*100 for i in sorted(pca.e_vals_, reverse=True)]
cum_var_exp = np.cumsum(var_exp)

with plt.style.context('seaborn-whitegrid'):
    fig, ax = plt.subplots(figsize=(6, 4))
    plt.bar(range(4), var_exp, alpha=0.5, align='center',
            label='individual explained variance')
    plt.step(range(4), cum_var_exp, where='mid',
             label='cumulative explained variance')
    plt.ylabel('Explained variance ratio')
    plt.xlabel('Principal components')
    plt.xticks(range(4))
    ax.set_xticklabels(np.arange(1, X.shape[1] + 1))
    plt.legend(loc='best')
    plt.tight_layout()
```

Unknown

and

```python
from sklearn.decomposition import PCA


pca = PCA(n_components=None)
pca.fit(X)
X_pca = pca.transform(X)

tot = sum(pca.explained_variance_)
var_exp = [(i / tot)*100 for i in sorted(pca.explained_variance_, reverse=True)]
cum_var_exp = np.cumsum(var_exp)

with plt.style.context('seaborn-whitegrid'):
    fig, ax = plt.subplots(figsize=(6, 4))
    plt.bar(range(4), var_exp, alpha=0.5, align='center',
            label='individual explained variance')
    plt.step(range(4), cum_var_exp, where='mid',
             label='cumulative explained variance')
    plt.ylabel('Explained variance ratio')
    plt.xlabel('Principal components')
    plt.xticks(range(4))
    ax.set_xticklabels(np.arange(1, X.shape[1] + 1))
    plt.legend(loc='best')
    plt.tight_layout()
```

Unknown


> Maybe I need to tweak a little the output or provide a slightly different input.

I think you are right though that the eigenvalues are not normalized in the mlxtend version  by default:

```python
pca = PrincipalComponentAnalysis(n_components=None)
pca.fit(X)
X_pca = pca.transform(X)
pca.e_vals_
```
```
array([2.91081808, 0.92122093, 0.14735328, 0.02060771])
```

(there was some reason why i wanted the raw eigenvalues, but I don't remember)

you could simply do that though via

```python
var_exp = [(i / tot)*100 for i in sorted(pca.explained_variance_, reverse=True)]
```

and see if it gives the same results then (I would be curious).

I can add

```python
var_exp = [(i / tot)*100 for i in sorted(pca.explained_variance_, reverse=True)]
```
also to the PCA class later as it might be useful in applications such as this (so that it doesn't have to be recomputed each time).

Btw should the arrow heads touch the circle or is the distance from the circle intended?

Gabriel-Azevedo-Ferreira: I ran your code and it seems that 
    - When doing `standardize(X)`, the results from both mlxtend and sklearn are the same
    - When not  doing `standardize(X)`, the results are actually different
    - **When not doing `standardize(X)`, but removing the mean from the mlxtend (and from sklearn if desired), the results are the same again**
 
`from mlxtend.data import iris_data
from mlxtend.preprocessing import standardize
from mlxtend.feature_extraction import PrincipalComponentAnalysis
import numpy as np

X, y = iris_data()

X_no_mean = X - X.mean(axis=0)
X_standardized = standardize(X)

print("mlxtend")
for X_,name in zip([X,X_no_mean,X_standardized],['X','X_no_mean','X_standardized']):
    print('\t',name)
    pca = PrincipalComponentAnalysis(n_components=None)
    pca.fit(X_)
    X_pca = pca.transform(X_)
    print(X_pca[:4,:])  

print("sklearn")
for X_,name in zip([X,X_no_mean,X_standardized],['X','X_no_mean','X_standardized']):
    print('\t',name)
    pca = PCA()
    pca.fit(X_)
    X_pca = pca.transform(X_)
    print(X_pca[:4,:])`


About the heads touching the circle, we can thing that it touches the outer part of the n-th dimensional circle. So for the projection in a plane, it is always inside the circle. 

The sum of the squares of the correlations is 1, because of the ortogonality of the PCA transformation matrix.

rasbt: > When not doing standardize(X), but removing the mean from the mlxtend (and from sklearn if desired), the results are the same again

oh interesting. so they probably do automatic mean-centering in their code

rasbt:
I think the failing unit tests could be because a new scikit-learn version was just recently uploaded to conda ... I will change the travis tests (here) to test on 0.20 and then carefully go through the 0.21 changes later

rasbt:
There were some odd things with the unit tests regarding nosetests; I think it's easier to get this all to work here instead of having to rebase after I do it in master. I switched to pytest now which doesn't seem to have issues with regard to the latest changes in sklearn for some of the things nosetests complained about.

rasbt:
Sorry for the mess, it was a bit of a coincidence that the new scikit-learn version was just pushed to conda when you created this PR so that this caused problems with some of the unit tests.

In any case, I made some minor adjustments to the documentation, and it should be all ok for merge now when the unit tests pass. Thanks a lot for the contribution!

Gabriel-Azevedo-Ferreira:
Thanks!
Glad I could contribute!

No problem and thank you for the discussions!

Issue #509, opened 2019-03-02
Feature Request - Frequent Patterns 


Sanjoth:
Suggestion to add **FP-Growth** and **Top-k Frequent itemsets** mining in frequent patterns library.

It would be great if mlxtend could have these 2 major algorithms which are also great for mining frequent patterns. It will be great if we can have all such tools in a single package.

Currently, the best FP-Growth implementation is in [Orange3](https://github.com/biolab/orange3-associate/blob/master/orangecontrib/associate/fpgrowth.py), but it doesn't have max_len which is a very big limitation when it comes to mining patterns in a large database.
As for top-k frequent itemsets mining, I didn't find any good library to use for production, it would be really great if these 2 can be added.

rasbt:
That's a good suggestion and I agree that this would be sth useful to have. There's an open issue about that -- currently "stale" though. I wouldn't have time to implement it myself at the moment but was hoping that someone else might be willing to help with a contribution here: #248.

harenbergsd:
I have an implementation of the FP-Growth algorithm (and the lesser known FPMax algorithm, which enumerates maximal itemsets), related to some research I was doing a while back.

I will look into into rewriting what I have in the style of this library and getting it added.

rasbt:
That would be amazing! The only real requirement/main consideration for the rewrite would be that it should have an interface similar to the current `apriori` implementation (http://rasbt.github.io/mlxtend/user_guide/frequent_patterns/apriori/) such that the output will be compatible with the `association_rules` function (http://rasbt.github.io/mlxtend/user_guide/frequent_patterns/association_rules/)

harenbergsd:
@Sanjoth do you have a problem statement for top-k itemsets? Is it, find the k highest supported itemsets that are at least some specified size?

harenbergsd:
Now that fp-growth is added, I can also add [FPMax](http://users.encs.concordia.ca/~grahne/papers/hpdm03.pdf) if you want @rasbt?

rasbt:
I’ve actually never heard of FPMax — I am not an expert in this area — but I certainly would appreciate it. The more the merrier ;)

Sanjoth:
> @Sanjoth do you have a problem statement for top-k itemsets? Is it, find the k highest supported itemsets that are at least some specified size?

[...]

Issue #248, opened 2017-09-17
Add Eclat and FPGrowth as alternatives to apriori for frequent itemset generation


rasbt:
Similar to 

    from mlxtend.frequent_patterns import apriori
    frequent_itemsets = apriori(df, min_support=0.6, use_colnames=True)

we could implement Eclat and FPGrowth as alternatives to apriori for frequent itemset generation. For instance,

    from mlxtend.frequent_patterns import eclat
    frequent_itemsets = eclat(df, min_support=0.6, use_colnames=True)

    from mlxtend.frequent_patterns import fpgrowth
    frequent_itemsets = fpgrowth(df, min_support=0.6, use_colnames=True)

The output should be consistent with apriori though so that the `frequent_itemsets` can be used as input to `mlxtend.frequent_patterns.association_r`

osama9374:
is eclat and fpgrowth algorithms added to mlxtend.frequent_patterns yet ? I tried using it. not working

rasbt:
Hi there,

no, eclat and fpgrowth have not been added to mlxtend, yet

rasbt:
@Demirrr tagging you here in case you'd like to continue the discussion regarding implementing eclat and/or fpgrowth

thomasSve:
How is the progress on this issue moving forward? I'm currently trying to build association rules for a large dataset and its super slow using apriori and was hoping to try out fpgrowth.

rasbt:
As I remember, there were some people interested in adding these but I don't think anyone had a chance to get to it yet. As for myself, I am all tangled up with teaching and other research projects so that I would not get a chance working on this myself anytime soon :(

tajuddinkhandaker:
> is eclat and fpgrowth algorithms added to mlxtend.frequent_patterns yet ? I tried using it. not working

---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-5-c653a61d2af5> in <module>()
----> 1 from mlxtend.frequent_patterns import fpgrowth

ImportError: cannot import name 'fpgrowth'

mjijas:
@rasbt I am trying implement eclat but this not yet included in mlxtend. 
what I am asking now, is there any way to implement eclat ?? using mlxtend 

```
frequent_itemsets = apriori(df, min_support=0.6, use_colnames=True)
rules = association_rules(frequent_itemsets, metric="support", support_only=True, min_threshold=0.1)
```

is this correct ??

[...]

PR #550, opened 2019-06-09
Add fpgrowth algorithm to frequent itemset package

harenbergsd:
### Description

Added FP-Growth algorithm for mining frequent itemsets.

Addresses issue #509 and #248

### Pull Request Checklist

A couple notes to discuss. 

(1) Regarding the documentation, there is a lot of overlap with apriori; same concept, same API, etc. Maybe we should have a generic documentation, that would basically be what 'Apriori' is now, except have it mention both algorithms and their APIs.

(2) Regarding the unit tests, I figured that we might as well reuse the same tests for both apriori and fpgrowth. I created a base class that defines the tests and then the tests for each algorithm subclass that base class and set a function object.


harenbergsd:
Hmm, I thought I didn't need to sync my fork, but it was out of date so now I have an extra merge commit here (master->feature branch). Let me know if you want me to close this and redo, so there isn't that ugly commit.

rasbt:
Thanks a lot for the PR! Regarding

> Let me know if you want me to close this and redo, so there isn't that ugly commit.

no need to worry about such things as "ugly commits" if everything seems to be complete. I haven't gone through all the code yet (currently traveling), but one issue I saw is that somehow the content of the mlxtend/mlxtend/frequent_patterns/tests/test_apriori.py file got deleted (and replaced by some setup code for `FPTestGrowth`, is that on purpose or an accident?

harenbergsd:
> one issue I saw is that somehow the content of the mlxtend/mlxtend/frequent_patterns/tests/test_apriori.py file got deleted (and replaced by some setup code for FPTestGrowth, is that on purpose or an accident?

Semi-intentional :). The class name was a typo, I meant to call it FPTestApriori, but otherwise yes, I meant to move most of the contents of that file into a different file (i.e., `test_fpbase.py`). 

This is related to what I was describing in (2) in the first comment. Because fpgrowth and apriori solve the same thing and have the same API it made sense to have shared unit test code. So, to use the same unit tests between them, I created a base class (FPTestBase) in `test_fpbase.py` and two subclasses (FPTestGrowth and FPTestApriori) in `test_fpgrowth.py` and `test_apriori.py`. The common unit test code (originally the code in `test_apriori.py`) went into FPTestBase.

I will wait to make any changes related to the unit tests until you let me know if you want to handle the unit test design for this in a different manner.

mlxtend/frequent_patterns/tests/test_fpbase.py
```diff
@@ -0,0 +1,123 @@
...
+                [0, 1, 0, 0, 0, 1, 1, 0, 0, 1, 1],
+                [0, 1, 0, 1, 1, 1, 0, 0, 1, 0, 0]])
+
+        self.cols = ['Apple', 'Corn', 'Dill', 'Eggs', 'Ice cream',
+                     'Kidney Beans', 'Milk',
+                     'Nutmeg', 'Onion', 'Unicorn', 'Yogurt']
+
+        self.df = pd.DataFrame(self.one_ary, columns=self.cols)
+
+        self.fpalgo = apriori
```
rasbt: this line is not really needed, right?

rasbt:
Oh I see now ... that is actually quite an elegant solution to avoid duplicating the tests and also making sure that both apriori and fpgrowth produce the same results.

What do you think about creating documentation? I'd suggest we make a standalone Jupyter Notebook like for apriori but then only add one simple, single example and mention that the usage is the same as for apriori and refer readers to see the apriori docs for more examples. We could then also add a benchmark example based on the numbers you quoted in the other "issue thread."

mlxtend/frequent_patterns/tests/test_fpbase.py
harenbergsd: Yeah, I think you are right.

However, if I remove it, it creates red error lines in my editor because it's not defined (but it will still work because it gets defined).

I guess the best way to handle it may be to force the argument in the constructor:

`test_fpbase.py`
```
    def setUp(self, fpalgo):
        self.one_ary = np.array(
            [[0, 0, 0, 1, 0, 1, 1, 1, 1, 0, 1],
                [0, 0, 1, 1, 0, 1, 0, 1, 1, 0, 1],
                [1, 0, 0, 1, 0, 1, 1, 0, 0, 0, 0],
                [0, 1, 0, 0, 0, 1, 1, 0, 0, 1, 1],
                [0, 1, 0, 1, 1, 1, 0, 0, 1, 0, 0]])

        self.cols = ['Apple', 'Corn', 'Dill', 'Eggs', 'Ice cream',
                     'Kidney Beans', 'Milk',
                     'Nutmeg', 'Onion', 'Unicorn', 'Yogurt']

        self.df = pd.DataFrame(self.one_ary, columns=self.cols)

        self.fpalgo = fpalgo
```

`test_fpgrowth.py`
```
class FPTestGrowth(unittest.TestCase, FPTestBase):
    def setUp(self):
        FPTestBase.setUp(self, fpgrowth)
```

`test_apriori.py`
```
class FPTestApriori(unittest.TestCase, FPTestBase):
    def setUp(self):
        FPTestBase.setUp(self, apriori)
```

harenbergsd:
That documentation plan sounds good to me 👍

rasbt:
Alright, just added the docs. Once the unit tests pass, I guess it's good to merge.  This really is a great PR, and I really appreciate it!

harenbergsd:
Cool, looks good! Thanks for doing the doc.

PR #553, opened 2019-06-18
Fpmax

harenbergsd:
### Description

Added [FPMax algorithm](http://users.encs.concordia.ca/~grahne/papers/hpdm03.pdf) to frequent pattern mining. FPMax find _maximal_ itemsets.

Related to #509

### Pull Request Checklist


mlxtend/frequent_patterns/apriori.py
```diff
@@ -105,13 +105,12 @@ def apriori(df, min_support=0.5, use_colnames=False, max_len=None, verbose=0):
...
 
     """
-    allowed_val = {0, 1, True, False}
-    unique_val = np.unique(df.values.ravel())
-    for val in unique_val:
-        if val not in allowed_val:
-            s = ('The allowed values for a DataFrame'
-                 ' are True, False, 0, 1. Found value %s' % (val))
-            raise ValueError(s)
+    idxs = np.where((df.values != 1) & (df.values != 0))
```
rasbt: Do you know if that captures `True` and `False` values as well by chance?

rasbt:
Again, thanks so much for this very nice PR!

I can take care of the documentation part.

mlxtend/frequent_patterns/apriori.py
harenbergsd: Yeah, in python, 1 == True (==1.0)

harenbergsd:
Ok, cool, thanks!

harenbergsd:
BTW, one design decision to discuss regarding max_len. It's not quite so clear what to do with max_len for this problem, where you are enumerating _maximal_ itemsets. There are two options:
1.  Given the set of all maximal itemsets, return those that are less than max_len
2.  Given the set of all itemsets less than max_len, return those that are maximal

The latter is including max_len in the maximality constraint. 

For example, if {a,b,c,d} and {c,d,e} are maximal itemsets. If you user gives a max_len of 3 the options would return:
1.  {c,d,e}
2.  {a,b,c}, {a,b,d}, {a,c,d}, {b,c,d}, {c,d,e}

I chose to go with the former, option (1), because with option (2) you can return itemsets that are not maximal wrt to the complete database. In other words, option (1) gives you subset of the complete set of solutions.

The other way would be fairly easy to implement as well, so if you think it's better, we can do it.

mlxtend/frequent_patterns/apriori.py
rasbt: haha, right, seems like my basic Python skill are getting a tad rusty

rasbt:
Good point. I think there would be edge cases where someone would prefer one over the other. However, when I think of maximal itemsets, intuitively, I would think about it wrt the complete database as in scenario 1. I'd say we should go with 1, like you also suggest.

rasbt:
Regarding what's currently implemented ... looking at the code (https://github.com/rasbt/mlxtend/blob/9be453c35f6642a6a9cd9ef57619702e246ed765/mlxtend/frequent_patterns/fpmax.py#L86), it looks like it's option 1, correct?

harenbergsd:
Yep, option 1.

rasbt:
Once the unit tests pass, I think this should be good to merge. Thanks a lot once again. Looks like quite some effort went into the code, and I particularly like the engineering best-practices regarding refactoring and encapsulation.

PR #555, opened 2019-06-19
possible to change the figure heigh and width

kazyka:
### Pull Request Checklist


mlxtend/plotting/plot_sequential_feature_selection.py
```diff
@@ -10,6 +10,7 @@
 
 
 def plot_sequential_feature_selection(metric_dict,
+                                      figsize=None
```
rasbt: I think you are missing a comma here after "None"

mlxtend/plotting/plot_sequential_feature_selection.py
```diff
@@ -50,12 +53,17 @@ def plot_sequential_feature_selection(metric_dict,
     http://rasbt.github.io/mlxtend/user_guide/plotting/plot_sequential_feature_selection/
 
     """
-
+    
+    if figsize is None:
+        figsize = (10,10)
```
rasbt: Hm, I think it would be better not to choose a particular figsize if it is `None`, otherwise it will be confusing to users. I suggest 

```python
if figsize is not None:
    fig = plt.subplots(figsize=figsize)
else:
    fig = plt.subplots()
```

mlxtend/plotting/plot_sequential_feature_selection.py
kazyka: Fixed it

mlxtend/plotting/plot_sequential_feature_selection.py
kazyka: I agree, it should be like this now

mlxtend/plotting/plot_sequential_feature_selection.py
```diff
@@ -23,7 +23,7 @@ def plot_sequential_feature_selection(metric_dict,
     Parameters
     ----------
     metric_dict : mlxtend.SequentialFeatureSelector.get_metric_dict() object
-    figsize : tuple (default: (10,10))
+    figsize : tuple (default: (8,6))
```
rasbt: Based on the function definition above, this should be

    figsize : tuple (default: None)

I think choosing `None` is also reasonable as it would use the matplotlib defaults

rasbt:
Thanks for the fix, and the PR! I addressed the documentation-specific points and it should be good to merge now.

kazyka:
Awesome, sorry for the sloppy edits, I have a deadline with my master thesis  due in a couple weeks. I have closed my comment. Hopefully I did it correct :)

PR #567, opened 2019-07-18
Apriori optimization

### LINKED ISSUE #566: Optimization of apriori algorithm by replacing iterator with matrix operations
Currently, the apriori implementation constructs a set of possible item combinations then iterates through this set to determine which combinations have a support above `min_support`. This iteration is slow and can be replaced by matrix equality operations and/or arithmetic. I propose the following changes:

```
    while max_itemset and max_itemset < (max_len or float('inf')):
        next_max_itemset = max_itemset + 1
        combin = np.array(list(generate_new_combinations(itemset_dict[max_itemset])))

        if combin.size == 0:
            break

        if verbose:
            print('\rProcessing %d combinations | Sampling itemset size %d' %
                  (combin.size[0], next_max_itemset), end="")

        if is_sparse:
            all_ones = np.ones((int(rows_count), 1))
            _bools = X[:, combin[:, 0]] == all_ones
            for n in range(1, combin.shape[1]):
                _bools = _bools & (X[:, combin[:, n]] == all_ones)
        else:
            _bools = np.all(X[:, combin], axis=2)

        support = _support(np.array(_bools), rows_count, is_sparse)
        _mask = (support >= min_support).reshape(-1)

        if any(_mask):
            itemset_dict[next_max_itemset] = np.array(combin[_mask])
            support_dict[next_max_itemset] = np.array(support[_mask])
            max_itemset = next_max_itemset
        else:
            break
```

In practice, this implementation is much faster than the current implementation, albeit at the cost of a slightly increased memory footprint: 

https://gist.github.com/jmayse/ad688d6a7fd842269996a701d7cecd4c

However, it usually remains slower than the `fpgrowth` implementation, as is expected:

https://gist.github.com/jmayse/7c76a2d838ac164b923a47b29527f2ed

- [x] Open a new "issue" on GitHub to discuss the new feature / bug fix
- [x] Fork the mlxtend repository from GitHub (if not already done earlier)
- [x] Create and check out a new topic branch (please don't make modifications in the master branch)
- [x] Implement the new feature or apply the bug-fix
- [ ] Add appropriate unit test functions in mlxtend/*/tests
- [x] Run nosetests ./mlxtend -sv and make sure that all unit tests pass
- [x] Check/improve the test coverage by running nosetests ./mlxtend --with-coverage
- [x] Check for style issues by running flake8 ./mlxtend (you may want to run nosetests again after you made modifications to the code)
- [ ] Add a note about the modification/contribution to the ./docs/sources/changelog.md file
- [ ] Modify documentation in the appropriate location under mlxtend/docs/sources/
- [x] Push the topic branch to the server and create a pull request
- [x] Check the Travis-CI build passed at https://travis-ci.org/rasbt/mlxtend
- [x] Check/improve the unit test coverage at https://coveralls.io/github/rasbt/mlxtend
- [ ] Check/improve the code health at https://landscape.io/github/rasbt/mlxtend

jmayse:
### Description

Currently, the implementation of the apriori algorithm uses a slow iteration to examine each item combination for above-threshold support. This iteration can be replaced by matrix operations that are generally faster, but use slightly more memory in some cases. 

While still generally slower than `fpgrowth`, this implementation of apriori is 3-6x faster: 

https://gist.github.com/jmayse/ad688d6a7fd842269996a701d7cecd4c

https://gist.github.com/jmayse/7c76a2d838ac164b923a47b29527f2ed

The current use of the `verbose` operator doesn't really make sense without the iterator; hence, I have replaced it with a statement that outputs the current number of combinations being compared. 

The exit behavior for the main `while` loop has also changed from setting `max_itemset = 0` to using `break`. If this is undesirable, we can easily reverse this change. 

Fixes #566

### Pull Request Checklist


rasbt:
Thanks a lot for the PR! 3-6x is definitely a substantial improvement and very welcome :). Regarding the implementation, it'll probably be more memory hungry since everything is done within an array instead of iterating over a generator, but I suppose it's not an issue on big datasets based on your experience with this code? (I assume, big datasets where this would be prohibitive would be so large that the iterative version would run ~forever anyways).

Otherwise, the alternative would be to enable both options via a flag "`in_memory=False`" (old version) and "`in_memory=True`" (new version). But I guess that's overkill!?

jmayse:
I don't really think that is overkill at all. The user should be able to specify a lightweight option in terms of memory uses for applications on low memory platforms (raspberry pi, for instance). Additionally the previous method using the iterator will always work - if given infinite time. There's value in that on its own. I'll add it back in under a flag with testing around the flag behavior.

rasbt:
> I don't really think that is overkill at all. The user should be able to specify a lightweight option in terms of memory usage for applications on low memory platforms (raspberry pi, for instance). Additionally the previous method using the iterator will always work - if given infinite time. There's value in that on its own. I'll add it back in under a flag with testing around the flag behavior.

Sounds good then. Also, the low-resource env makes sense. Thanks!

mlxtend/frequent_patterns/tests/test_fpbase.py
```diff
@@ -133,6 +147,20 @@ def test_max_len(self):
...
         assert max_len == 2
 
+    def test_low_memory_flag(self):
+        try:
+            with captured_output() as (out, err):
+                _ = self.fpalgo(self.df, low_memory=True, verbose=1)
+                print(out.getvalue())
+        except TypeError:
+            # If there is no low_memory argument, don't run the test.
+            assert True
```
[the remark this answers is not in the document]
rasbt: Hm, since it's a function, that's a tricky one ... one solution, which is not that much better, would be

```
import inspect

    def test_low_memory_flag(self):

        if 'low_memory' in inspect.getargspec(self.fpalgo)[0]:
             with captured_output() as (out, err):
                 _ = self.fpalgo(self.df, low_memory=True, verbose=1)
                 print(out.getvalue())
         else:
             # If there is no low_memory argument, don't run the test.
             assert True
```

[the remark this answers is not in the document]
jmayse: Actually, I like that better. I've changed this test to use `inspect.signature` and removed that pesky debug `print` that snuck through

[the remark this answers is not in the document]
rasbt: sounds good!

mlxtend/frequent_patterns/apriori.py
```diff
@@ -122,50 +141,80 @@ def apriori(df, min_support=0.5, use_colnames=False, max_len=None, verbose=0):
                              '`df.columns = [str(i) for i in df.columns`].')
 
         X = df.to_coo().tocsc()
-        support = np.array(np.sum(X, axis=0) / float(X.shape[0])).reshape(-1)
+        support = _support(X, X.shape[0], is_sparse)
     else:
         X = df.values
-        support = (np.sum(X, axis=0) / float(X.shape[0]))
+        support = _support(X, X.shape[0], is_sparse)
```
rasbt: Here, 

```python
        X = df.to_coo().tocsc()
         support = np.array(np.sum(X, axis=0) / float(X.shape[0])).reshape(-1)		         support = _support(X, X.shape[0], is_sparse)
     else:		     else:
         X = df.values		         X = df.values
         support = (np.sum(X, axis=0) / float(X.shape[0]))		         support = _support(X, X.shape[0], is_sparse)

``` 
the line

```python
support = _support(X, X.shape[0], is_sparse)
```

could be put after the if-else clause right? It looks identical and would be duplicated otherwise

mlxtend/frequent_patterns/apriori.py
```diff
@@ -122,50 +141,80 @@ def apriori(df, min_support=0.5, use_colnames=False, max_len=None, verbose=0):
...
+
+        if verbose and not low_memory:
+            print('\rProcessing %d combinations | Sampling itemset size %d' %
+                  (combin.size, next_max_itemset), end="")
+
+        # With exceptionally large datasets, the matrix operations can use a
+        # substantial amount of memory. For low memory applications or large
+        # datasets, set low_memory = True to use a slower but more memory-
+        # efficient implementation.
+        if low_memory:
```
rasbt: I think the code would be a little bit more efficient if 

```python
if not low_memory:
...
if low_memory:
```

was changed to

```python
if not low_memory:
...
else:
```

jmayse: Agreed, that was sloppy. Fixed!

rasbt: thanks!

mlxtend/frequent_patterns/apriori.py
```diff
@@ -105,6 +111,19 @@ def apriori(df, min_support=0.5, use_colnames=False, max_len=None, verbose=0):
     http://rasbt.github.io/mlxtend/user_guide/frequent_patterns/apriori/
 
     """
+
+    def _support(_x, _n_rows, _is_sparse):
```
rasbt: Not really important since private functions don't show up in the docs, but usually, we use the NumPy-style documentation style. That's because I had written a Python doc to markdown parser that automatically generates the documentation for the HTML website documentation. No need to change it here though because like I said, it won't appear on the website

jmayse: Well, this is definitely the time to change it if you want me to. I'm happy to do it to conform to the overall style.

jmayse: In fact, if you don't mind, let me just do that real quick...even if it doesn't appear on the website we should be consistent to the overall style. Why build in inconsistencies?

jmayse: Done

rasbt: ok sure, please go ahead then. Thanks!

rasbt:
I think everything should be good to go now (once the tests pass). Just added an additiona tests such that the whole suite runs for both `low_memory=True` & `low_memory=False`. Besides that, I also just added the Changelog entry. 

Unless you have further suggestions, I would merge it once the tests pass. Will then also make a new version release later tonight or tomorrow -- it's about time ;).

Thanks a lot for this PR!

jmayse:
Happy to! Thanks for maintaining the library!

Issue #593, opened 2019-09-23
Plot correlation graph bug + feature


Gabriel-Azevedo-Ferreira:
The Plot correlation graph is a bit slow, because I initially used `np.corrcoef` to create the correlation matrix, which add extra complexity.

Also, I noted a bug on the explaned variance plotting (the percentages didn't sum to one).

I fixed those issues and will pull request soon! :-)

jbbqqf:
Hi! Triaging older issues this week — this is a 6-year-old issue where the reporter said "I fixed those issues and will pull request soon!" but no PR ever followed.

Checking the current `master`:

- The "slow because of `np.corrcoef`" point: [`mlxtend/plotting/pca_correlation_graph.py:43-45`](https://github.com/rasbt/mlxtend/blob/master/mlxtend/plotting/pca_correlation_graph.py#L43-L45) shows the old `np.corrcoef` line is now commented out and replaced by a `corr2_coeff(A.T, B.T).T` call, so the performance improvement seems to have made it in via another path.
- The "explained variance percentages didn't sum to one" point: the function now uses `mlxtend.feature_extraction.PrincipalComponentAnalysis`'s `e_vals_` for the variance bar plot (see lines around the `pyplot.bar(..., explained_variance / explained_variance.sum() * 100, ...)` block), which is normalised by construction.

So both of the original symptoms appear to be addressed in the current code. Without a fresh reproducer on a recent version this looks safe to close — happy to be wrong if there's still a divergence I missed.

(Disclosure: I drafted this comment with help from Claude Code while triaging stale issues; the file/line references above were verified manually against the current `master`.)

rasbt:
Just some housekeeping to clear out the issues. Yes, this has long been addressed :)

PR #594, opened 2019-09-23
fast corr, solving bug on expl. var, removing pca args 

Gabriel-Azevedo-Ferreira:
### Description

Related to issue:
[#593](https://github.com/rasbt/mlxtend/issues/593)

### Pull Request Checklist


mlxtend/plotting/pca_correlation_graph.py
```diff
@@ -55,10 +75,9 @@ def plot_pca_correlation_graph(X, variables_names, dimensions=(1, 2),
     X = X - X.mean(axis=0)
     n_comp = max(dimensions)
 
-    if X_pca is None:
-        pca = PrincipalComponentAnalysis(n_components=n_comp)
-        pca.fit(X)
-        X_pca = pca.transform(X)
+    pca = PrincipalComponentAnalysis(n_components=n_comp)
```
rasbt: In the previous version, we allowed users to provide their own PC axes as input to the plot. In this updated version, we know have to run PCA on the dataset for producing the plot. I think this simplifies the use/API in general and would generally be ok, but I am curious: what would be the reason for this change?

rasbt:
Thanks a lot for noting the issues and updating this function. Let's continue the discussion from #594 here then.

rasbt:
PS: there was no changelog entry yet, so I "unchecked" the boxes above (no worries about that for now)

rasbt:
Thanks for addign the changelog! Have you seen my comment above regarding the removal of the PCA args -- just curious what motivated this.

Gabriel-Azevedo-Ferreira:
I did! Sorry, did not have time to reply earlier.
I had forgotten to change the change log. I can change the notebook too later and commit. Or you prefer to change yourself?

I think it can create some possibility of bugs if we allow the PCA to be given by the user. And we would also would have to tackle the case where the user asks to plot dimension X and the PCs matrix returns  over less than X dimension...

Gabriel-Azevedo-Ferreira:
I think I closed the pull request by mistake 😳
What should I do? create a new one?

rasbt:
> I can change the notebook too later. Or do you prefer to change it yourself?

No worries about the notebook. I can take care of that.

> I think it can create some possibility of bugs if we allow the PCA to be given by the user (e.g, the algorithm the user is using gives the components in different order or the matrix is transposed). And we would also would have to handle the case where the user asks to plot dimension X and the PCA returns less than X components (the default for sklearn is 5 I think).

Generally, I agree with you -- weird things can happen if users provide their own data. On the other hand, we could simply 

# 1) 

specify in the docstring for the input `X` that we expect the matrix to be an `np.ndarray, shape = [n_samples, n_components]`

# 2)

also just add a check like

```python
if `n_comp` < X.shape[1]:
    raise ValueError(f"Input array `X` contains fewer principal"
                      " components than expected based on `dimensions`."
                      " Got {X.shape[1]} components in X,"
                      " expected `max(dimensions)={n_components}.")
```

What do you think?

Gabriel-Azevedo-Ferreira:
I agree! I was doing the modifications and noticed that we also needed the eigenvalues (to put the explained variance of each component in the same plot).

So I think I should either let the user provide both (eigenvalues and eigenvectors) as inputs, or neither. 

I think it doesnt hurt to provide both...

Gabriel-Azevedo-Ferreira:
I agree!
while doing the test cases, I noticed that I also need to add an argument for the explained variance.

I added that argument as well as your suggestions for the checks and docstring.

Btw, the explained variance from the mlxtend version (`e_vals_`) is slightly different from the scikit learn version.

In sklearn, they take the variance (the eigenvalues) dividing by `n-1` while in mlxtend it is divided by `n`:

https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/decomposition/pca.py line 452
```
        # Get variance explained by singular values
        explained_variance_ = (S ** 2) / (n_samples - 1)
        total_var = explained_variance_.sum()
        explained_variance_ratio_ = explained_variance_ / total_var
        singular_values_ = S.copy()  # Store the singular values.
```

This makes a small difference, because, when I compute the total variance (the sum of the variances on the original data) and divide by `n`, which is coherent with the mlxtend version but not the sklearn version.

what do you suggest me to do?

rasbt:
> In sklearn, they take the variance (the eigenvalues) dividing by n-1 while in mlxtend it is divided by n:

Oh, good catch! using *n-1* is actually the right thing to do from a stats perspective. NumPy really is a confusing tool though :D. By default, they use *1/n* for `np.var` but *n-1* degrees of freedom in `np.cov`.

I just noticed that the unit tests didn't have enough precision to catch this discrepency between the `eigen` and `svd` mode. I am fixing this now via #595. Thanks for noticing that! Please use *n-1*.

rasbt:
Also, I am not sure why the one test on iris suddenly fails. I am looking into it.

rasbt:
Alright, just fixed these Iris issues + the $n-1$ scaling in the `svd`-mode PCA. Maybe try to do a rebase

Gabriel-Azevedo-Ferreira:
Ok! great then. just rebased, changed to `n-1`, did a small change on the test case and pushed again.

-----

Also, the tests have failed, I  think the reason may be because I call the iris db outside of the test function, like this:

```
X, y = iris_data()
def test_pass_pca_corr():
    plot_pca_correlation_graph(X, ['1', '2', '3', '4'])
def test_pass_pca_corr_pca_out():
    ...
```
I changed this and pushed again, let's see if the problem is solved (sorry for the trouble anyway!)

rasbt:
Looks good, thanks! I added a few more tests.

Gabriel-Azevedo-Ferreira:
Great! Thanks!

rasbt:
it should be good to merge then as an update to the previous function, unless you have further suggestions

Gabriel-Azevedo-Ferreira:
I am good. Thanks

And sorry for error in the iris test, which was actually caused by my test
csse.

rasbt:
No worries, and thanks for the this PR and the fix!

Issue #615, opened 2019-11-01
per-class accuracy 


ghayth82:
hi
per-class accuracy is provide the average per-class accuracy , i suggest to add per-class accuracy . I mean to calculate accuracy of each class individually, so for example  to check which class affecting the model and has the lowest accuracy

rasbt:
I think that might be useful in certain applications. It should probably be a standalone function `per_class_accuracy` (similar to `accuracy_score` in https://scikit-learn.org/stable/modules/generated/sklearn.metrics.accuracy_score.html) so that users can choose which class to use as the "positive class). I.e., adding it to the current `scoring` function wouldn't work, because it wouldn't allow that choice.

rasbt:
Fixed in #624

PR #624, opened 2019-11-08
Created new per class accuracy function.

### LINKED ISSUE #617: Standalone `average_per_class_accuracy` function
Make a standalone `average_per_class_accuracy_score` function akin to https://scikit-learn.org/stable/modules/generated/sklearn.metrics.accuracy_score.html

deepandas11:
Features:
* new pos_label argument(got to change name, as there could be ambiguity
with positive_label in the original scoring function). Is None by
default, in which case, average accuracy is returned
* uses existing _accuracy function for computing per class accuracy

### Description

 
This PR intends to introduce a standalone per class accuracy function 

Fixes  #617 

### Pull Request Checklist


mlxtend/evaluate/scoring.py
```diff
@@ -111,10 +120,10 @@ def scoring(y_target, y_predicted, metric='error',
     elif metric == 'error':
         res = _error(targ_tmp, pred_tmp)
     elif metric == 'average per-class accuracy':
```
rasbt: I think there was a misunderstanding. The per-class accuracy would not be used to replace the average-per-class-accuracy. Maybe have a look at the brief comment in #615

deepandas11: Yes! This makes much more sense now. Thanks!

deepandas11:
Hi, @rasbt !
Please let me know if there are any corrections/recommendations!

mlxtend/evaluate/per_class_accuracy.py
```diff
@@ -0,0 +1,43 @@
...
+    Parameters
+    ------------
+    y_target : array-like, shape=[n_values]
+        True class labels or target values.
+    y_predicted : array-like, shape=[n_values]
+        Predicted class labels or target values.
+    pos_label : str or int, None by default.
+        The class whose accuracy score is to be reported.
+        'binary' has to be set to True.
+    binary : bool
```
rasbt: What happens if `binary` is set to `False`. Will this be a regular accuracy computation then? I think it looks like that. It would be nice to document that. 

I have another suggestion, though. How about 

1) calling this function just "accuracy" and computing the "average" per-class accuracy as an option?

E.g., instead of `binary`, we could use another parameter called `method`.

- `method='standard'` computes the regular accuracy
- `method='binary'` computes per-class accuracy for the positive class
- `method='average'` computes average per class accuracy similar to https://github.com/rasbt/mlxtend/blob/master/mlxtend/evaluate/scoring.py#L113

What do you think?

rasbt:
Sorry, it was a busy week and just got a chance to look at it. I have a suggestion (mention in the comment above). Let me know what you think. Also, it's weird that the unit test for Windows fails. Not sure if this is necessarily related to this PR or a general issue. Will investigate.

rasbt:
Just did a rebase of your changes on top of the master branch to resolve the CI issues. You would need to do a 

    git pull origin per_class_accuracy

on your end before making any additional changes.

mlxtend/evaluate/per_class_accuracy.py
deepandas11: Yes, that should have been documented. Thanks for pointing that out.

As for the suggestion, I think if you are looking to introduce modularity with respect to the individual metrics in "scoring.py", this would be a good first step. 
However, it would be interesting to know if you indeed want to eventually deprecate the current methodology of finding metrics and move towards standalone functions, like this?

deepandas11:
Thank you so much for resolving the CI issue.
No worries about the delay.

mlxtend/evaluate/per_class_accuracy.py
rasbt: I think it would be good to support both, like scikit-learn does. The 'string' arguments in scoring are useful as a short-cut. However, it doesn't allow options like `pos_label`. Inside scoring, the idea would then be to use the `per_class_accuracy` function to implement the current `'average per-class accuracy'`

deepandas11: So, implementation-wise, I think the API would be less confusing if we have just a single accuracy function that can compute the three types of metrics, based on the choice of "method". This however, means no separate per_class_accuracy method, as was originally intended.

rasbt: I agree. I think it's best if we call it just `classification_accuracy`? And then inside we have the `"per class"`  option

rasbt: i have to correct this, "`accuracy_score`" may be better because we also have `lift_score` already, and it is generally more consistent with sklearn then.

deepandas11:
So, I have made the changes as discussed. 
Another feature we can now possibly add to the accuracy scores is the `normalize` feature like in sklearn. (But maybe in a separate PR. I can open an issue if you feel we should add it)

mlxtend/evaluate/accuracy.py
```diff
@@ -0,0 +1,57 @@
...
+# mlxtend Machine Learning Library Extensions
+#
+# A function for generating per class accuracy
+#
+# License: BSD 3 clause
+import numpy as np
+
+
+def _compute_metric(target_temp, predicted_temp):
+    return (target_temp == predicted_temp).sum() / float(target_temp.shape[0])
```
rasbt: This line,

    return (target_temp == predicted_temp).sum() / float(target_temp.shape[0])

could be even more simplified to

    return (target_temp == predicted_temp).mean()

I saw that you suggested the `normalize`. Since it is a simple thing, we could already add it here instead of a separate PR. When I understand this correctly, this would simply be

```python
def _compute_metric(target_temp, predicted_temp, normalize=True):
    if normalize:
        return (target_temp == predicted_temp).mean()
    else:
        return (target_temp == predicted_temp).sum()
```

mlxtend/evaluate/accuracy.py
```diff
@@ -0,0 +1,57 @@
...
+        Predicted class labels or target values.
+    method : str, 'standard' by default.
+        The chosen method for accuracy computation.
+        If set to 'standard', computes overall accuracy.
+        If set to 'binary', computes accuracy for class pos_label.
+        If set to 'average', computes average per class accuracy.
+    pos_label : str or int, 1 by default.
+        The class whose accuracy score is to be reported.
+        Used only when method is set to 'binary'
+
```
rasbt: Add doc for `normalize`

mlxtend/evaluate/accuracy.py
```diff
@@ -0,0 +1,57 @@
...
+    y_predicted : array-like, shape=[n_values]
+        Predicted class labels or target values.
+    method : str, 'standard' by default.
+        The chosen method for accuracy computation.
+        If set to 'standard', computes overall accuracy.
+        If set to 'binary', computes accuracy for class pos_label.
+        If set to 'average', computes average per class accuracy.
+    pos_label : str or int, 1 by default.
+        The class whose accuracy score is to be reported.
+        Used only when method is set to 'binary'
```
rasbt: I suggest changing

    Used only when method is set to 'binary'

to 

    Used only when `method` is set to 'binary'

deepandas11:
Have made the changes. Although, I think you have already seen that I made a few changes to `scoring.py` already to accommodate the new accuracy function.

rasbt:
Thanks a lot! I am currently a bit swamped but will try to take a look at it on the weekend!

mlxtend/evaluate/accuracy.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def _compute_metric(target_temp, predicted_temp, normalize=True):
+    if normalize:
+        return (target_temp == predicted_temp).mean()
+    else:
+        return (target_temp == predicted_temp).sum()
+
+
+def accuracy_score(y_target, y_predicted, method="binary",
```
rasbt: I think we want to change that to `method="standard"` as default value.

mlxtend/evaluate/accuracy.py
```diff
@@ -0,0 +1,64 @@
...
+
+    elif method == "binary":
+        if pos_label not in unique_labels:
+            raise AttributeError("Chosen value of pos_label doesn't exist")
+
+        target_temp = np.where(target_temp == pos_label, 1, 0)
+        predicted_temp = np.where(predicted_temp == pos_label, 1, 0)
+        return _compute_metric(target_temp, predicted_temp, normalize)
+
+    else:
```
rasbt: To make this function a bit less error prone, I would suggest changing the current `else` to 

    elif method == "average":

and then add an additional 

    else:
        raise ValueError('`method` must be "standard",
                         "binary", or "average". Got %s.' % method)

deepandas11: Makes perfect sense. Updated everything.

rasbt:
Thanks for making those changes. Looks good to me now. I added the documentation, let me know if that looks good to you.

deepandas11:
Looks great.

rasbt:
Cool thanks! I will merge then!

PR #650, opened 2019-12-21
15th attempt

lareinaxy:
### Pull Request Checklist

### DIFF
--- a/mlxtend/feature_selection/sequential_feature_selector.py
+++ b/mlxtend/feature_selection/sequential_feature_selector.py
@@ -382,7 +382,8 @@ class SequentialFeatureSelector(_BaseXComposition, MetaEstimatorMixin):
 
         else:
             select_in_range = False
-            k_to_select = self.k_features
+            k_to_select = int(len(X[1])**.5) # random subsample of sqrt(k_features) dimensions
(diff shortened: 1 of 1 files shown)


mlxtend/feature_selection/sequential_feature_selector.py
```diff
@@ -382,7 +382,8 @@ def fit(self, X, y, custom_feature_names=None, groups=None, **fit_params):
 
         else:
             select_in_range = False
-            k_to_select = self.k_features
+            k_to_select = int(len(X[1])**.5)
+            np.take(X, np.random.permutation(X.shape[1]), axis=1, out=X)
```
rasbt: Hi Lareina, thanks for the PR. I don't quite understand how the value is chosen `k_to_select = int(len(X[1])**.5)`. So this means that the random subset is always sqrt of the original number of features?

There are two problems with that.

1) We want to add the random feature subset size as an optional thing
2) I think we should let the user allow to choose the subset size

There could be a new parameter 

`use_random_feature_subset` for the `SequentialFeatureSelector` class which either accepts a function like `f = lambda x: int(np.sqrt(x)` or `None`.

Let me know if you have questions.

PR #683, opened 2020-04-20
Fix typo in `paired_ttest_kfold_cv` docs

bryant1410:


rasbt:
Good catch, thanks for the fix!

PR #705, opened 2020-07-14
Forward slection reduce mem copy

### LINKED ISSUE #704: Out of memory problem with sequential forward selection
Hey :-)

I have a problem where my computer runs out of memory, when using the Sequential Feature *Forward* Selection.
I have a moderate sized dataset of roughly 2gb of data, but over 2000 features I am selecting from on all cores of the machine.

The computer:
256gb ram
48 cores
Ubuntu 18.04
Python 3.6

As far as i can tell the problem is in the implementation of the parallel backend. It copies the full data to the subprocess and then does the slicing there. On a backward selection this should be find, but for forward only a partial of that data is actually needed. 
I would propose doing the slicing before calling the _calc_score function.

Saniamos:
### Description

Implements slicing the features from the data before copying it to the subprocess.

Fixes #704

### Pull Request Checklist

- [x ] Added a note about the modification or contribution to the `./docs/sources/CHANGELOG.md` file (if applicable)
- [ x] Ran `PYTHONPATH='.' pytest ./mlxtend -sv` and make sure that all unit tests pass (for small modifications, it might be sufficient to only run the specific test file, e.g., `PYTHONPATH='.' pytest ./mlxtend/classifier/tests/test_stacking_cv_classifier.py -sv`)
- [x ] Checked for style issues by running `flake8 ./mlxtend`

### DIFF
--- a/docs/sources/CHANGELOG.md
+++ b/docs/sources/CHANGELOG.md
@@ -25,6 +25,7 @@ The CHANGELOG for the current development version is available at
 
 - Implemented both `use_clones` and `fit_base_estimators` (previously `refit` in `EnsembleVoteClassifier`) for `EnsembleVoteClassifier` and `StackingClassifier`. ([#670](https://github.com/rasbt/mlxtend/pull/670) via [Katrina Ni](https://github.com/nilichen))
 - Switched to using raw strings for regex in `mlxtend.text` to prevent deprecation warning in Python 3.8 ([#688](https://github.com/rasbt/mlxtend/pull/688))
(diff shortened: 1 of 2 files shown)


rasbt:
Oh yeah, like you mentioned in #704, subprocesses require everything to be copied. You are right, we can slice the data before sending it to the processes. Thanks for implementing it!

Saniamos:
> Oh yeah, like you mentioned in #704, subprocesses require everything to be copied. You are right, we can slice the data before sending it to the processes. Thanks for implementing it!

sure, thanks for the quick merge :-)

PR #725, opened 2020-09-04
Keras Sequential Model Support for Bias Variance Computation

hanzigs:


rasbt:
Thanks, @hanzigs, this looks great. Ideally, we could add a unit test, but this would add yet another dependency to the package, so I think it's fine leaving it like this. Will merge shortly. Thanks!

hanzigs:
Thank you very much

PR #731, opened 2020-09-28
Fix small typo in Fpmax user guide

andrewm4894:
Change 

"However, the runtime of Apriori can be quite small" 
to 
"However, the runtime of Apriori can be quite large"

"small" to "large" as I think that might be how it was meant to originally read.


rasbt:
Good catch, thanks for the PR!

Issue #833, opened 2021-06-22
Custom generator for models in exhaustive feature selector


jonathan-taylor:
#### Describe the workflow you want to enable

I'd like to make it easier to do best subsets with categorical features -- for simplicity let's start by assuming an additive model so for each feature there are a set of columns in the design matrix associated with that feature. When all are continuous features
each feature is associated to a single column, otherwise there is a feature grouping that can be described as a sequence of length `X.shape[1]` assigning columns to a particular feature. More generally, this sequence assigning columns to features could also include interactions of both continuous and categorical variables.

#### Describe your proposed solution

It is (at least in some corners) common practice to include all columns associated to a categorical feature or none. This would be able to be encoded in the `candidates` list. If interactions were permitted then some conventions only include an interaction if both main effects are also included. While the logic of which candidates to generate may be user-specific, it would seem if we could supply a custom iterator for `candidates` then most of the code should not need to be modified. Instead of `custom_names` each particular candidate may have its own identifier, so one could specify
whether the iterator produces simply indices or (indices, identifier) pairs.

This would remove the need for the `min_features/max_features` argument as this would be encoded into the
iterator itself. So perhaps a helper functions to produce at least a few common iterators for candidates could be included.
Specifically one which produce the default "all continuous" iterator, and one which could easily handle an additive model
with possibly some categorical variables. 

#### Describe alternatives you've considered, if relevant

I've considered simply wrapping `R` functions like `regsubsets` that easily handles the categorical variables. I would
prefer an sklearn aware version that could do this as well.

jonathan-taylor:
Implemented a simple version here: https://github.com/rasbt/mlxtend/pull/834

It might also be nice to have the possibility of the sequential feature selector use custom logic as well. Again, when adding and deleting categorical variables or interactions one would want to add or delete groups of features at a time.

rasbt:
Overall, this sounds like a great idea, and I would be in favor of such a solution for both the exhaustive and sequential feature selectors. Refactoring this into custom iterators seems like a very elegant solution. We would then have helper function that reproduces the current behavior along with generating iterators for the datasets with categorical variables. 

With regard to identifying categorical features, there are many options, but what do you think of adopting the approach used in scikit-learn's HistGradientBoostingClassifier (https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.HistGradientBoostingClassifier.html)?

[...]

PR #834, opened 2021-06-22
WIP: having exhaustive feature selector take a custom iterator as input

jonathan-taylor:
### Code of Conduct

A working candidate for https://github.com/rasbt/mlxtend/issues/833

### Pull Request Checklist

### DIFF
--- a/mlxtend/feature_selection/exhaustive_feature_selector.py
+++ b/mlxtend/feature_selection/exhaustive_feature_selector.py
@@ -25,7 +25,7 @@ from sklearn.model_selection import cross_val_score
 from joblib import Parallel, delayed
 
 
-def _calc_score(selector, X, y, indices, groups=None, **fit_params):
+def _calc_score(selector, X, y, name, indices, groups=None, **fit_params):
     if selector.cv:
         scores = cross_val_score(selector.est_,
                                  X[:, indices], y,
@@ -38,7 +38,7 @@ def _calc_score(selector, X, y, indices, groups=None, **fit_params):
     else:
         selector.est_.fit(X[:, indices], y, **fit_params)
         scores = np.array([selector.scorer(selector.est_, X[:, indices], y)])
-    return indices, scores
+    return name, indices, scores
 
 
 def _get_featurenames(subsets_dict, feature_idx, custom_feature_names, X):
@@ -75,10 +75,6 @@ class ExhaustiveFeatureSelector(BaseEstimator, MetaEstimatorMixin):
     Parameters
     ----------
     estimator : scikit-learn classifier or regressor
-    min_features : int (default: 1)
-        Minumum number of features to select
-    max_features : int (default: 1)
-        Maximum number of features to select
     print_progress : bool (default: True)
         Prints progress as the number of epochs
         to stderr.
@@ -149,14 +145,12 @@ class ExhaustiveFeatureSelector(BaseEstimator, MetaEstimatorMixin):
     http://rasbt.github.io/mlxtend/user_guide/feature_selection/ExhaustiveFeatureSelector/
 
     """
-    def __init__(self, estimator, min_features=1, max_features=1,
+    def __init__(self, estimator, 
                  print_progress=True, scoring='accuracy',
                  cv=5, n_jobs=1,
                  pre_dispatch='2*n_jobs',
                  clone_estimator=True):
         self.estimator = estimator
-        self.min_features = min_features
-        self.max_features = max_features
         self.pre_dispatch = pre_dispatch
         self.scoring = scoring
         self.scorer = get_scorer(scoring)
@@ -176,7 +170,7 @@ class ExhaustiveFeatureSelector(BaseEstimator, MetaEstimatorMixin):
         # don't mess with this unless testing
         self._TESTING_INTERRUPT_MODE = False
 
-    def fit(self, X, y, custom_feature_names=None, groups=None, **fit_params):
+    def fit(self, X, y, candidates, custom_feature_names=None, groups=None, **fit_params):
         """Perform feature selection and learn model from training data.
 
         Parameters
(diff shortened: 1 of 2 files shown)


jonathan-taylor:
Obviously some linting needed, but have generalized the sequential / exhaustive feature selector to allow for categorical and custom search strategies. The previous searches default to special cases now. Post-processing of search results still need implementing.

jonathan-taylor:
The type checking is more just a declaration of what a strategy is at this point.

Will work on PEP8 issues and also will try to have EFS and SFS use the generic one. I think EFS is straightforward, but SFS I have to understand the "floating" logic a little better -- I gather this is just "search both forward and backward" but somehow in self.subsets_​ we only keep a given model of a fixed size -- is this meant to be the best model of a fixed size? Or maybe I still have the logic wrong.
________________________________

rasbt:
> I think EFS is straightforward, but SFS I have to understand the "floating" logic a little better -- I gather this is just "search both forward and backward" but somehow in self.subsets_​ we only keep a given model of a fixed size -- is this meant to be the best model of a fixed size? Or maybe I still have the logic wrong.

Oh, I should have described this a bit better in the docs. In the floating forward variant, it allows you to delete a previously selected feature if it improves performance. In the floating backward variant, it allows you to add back a previously deleted feature.

Maybe it is easier to explain with a toy example.

Say we have a feature set [0, 1, 3, 4, 5] and we do sequential "floating" forward selection. 

## Round 1

Starting with the empty set [], we check all features and find that feature 1 is best:

- Select feature 1 to be added to [ ]

- Now we have the optional floating step where we can optionally delete a previously selected feature. Here, this does not apply because we only have the currently selected feature.

- Return feature set [1]


## Round 2

Starting with the set [1], we check all features and find that feature 4 is best:

- Select feature 4 to be added to [1]

- Optional floating step. Here, we can try removing 1 but this can be skipped, because it would be the same as selecting the 4 in the first place in the previous round. We can try removing it for simplicity though.

- Return feature set [1, 4]

## Round 3

- Select feature 2 to be added to [1, 4].

- Optional floating step. Here, we can try removing 1 or 4. Removing 4 can technically be skipped, because it would be the same as selecting the 2 in the first place in the previous round. We can try removing it for simplicity though. I.e., we try [1, 2], [2, 4].

- return [1, 2, 4] (assuming it is better than [1, 2], [2, 4])



## Round 4

- Select feature 3 to be added to [1, 2, 4].

- Optional floating step. Here, we can try removing 1, 2, or 4.  We try [1, 2, 3], [1, 3, 4], [2, 3, 4].

- return [1, 2, 3] (assuming it is better than  [1, 3, 4], [2, 3, 4], [1, 2, 3, 4])


Note that we would not have selected [1, 2, 3] with regular forward selection. With regular forward selection, we would have had

[ ] -> [1] -> [1, 4] -> [1, 2, 4]

Or in other words, with regular forward selection, any feature set of length 3 would have included the value 4. However, in combination with other features, 4 may not be useful.

So, in practice, the floating forward variant allows us to test a few more combinations than the regular forward variant, without being as expensive as the exhaustive feature selection of course.

The floating backward variant works similar to the floating forward variant, except we are allowed to add a previously removed feature back.

jonathan-taylor:
So it seems floating is as I thought. This is "both" - the stepwise will look to add or drop columns from current set. This is what "both" in R's step is.

My issue with the logic was just somehow how you tracked the progress within self.subsets_ as there is no longer necessarily a single model of size 3, say. The generic one just stores every state visited.

Get Outlook for Android<https://aka.ms/ghei36>
________________________________

rasbt:
Yeah, I think with the new version it will be way easier to track.

Regarding "both" there is a difference between floating-forward and floating-backward though. They can both give you different end results. I think "both" is currently doing floating-forward?

rasbt:
Thanks for the updates. Actually, some of the failing tests are related to sklearn 1.0. Due to bug fixes or changes, their have been very slight changes in some of the accuracy values, which will cause values when comparing the results on 3 positions after the decimal point. I am currently looking into this and will update the tests

jonathan-taylor:
Got it. I am pretty comfortable with the state of the code, but I think more tests are needed on my end.

It is possible to rewrite the two current estimators in this form. Do you think that's a high priority?
________________________________

rasbt:
I am not sure how many people use the floating versions. One option could be to deprecate/remove them. We could maybe add a module `mlxtend.legacy` for that where we can keep the old sequential feature selection code for a while. However, I think think floating variants can be quite powerful. Just looking at the paper again (attached), here is a comparison of regular forward and backward selection compared to an exhaustive ("optimal") method:
Screen Shot 2021-11-02 at 1 52 21 PM
[1-s2.0-0167865594901279-main.pdf](https://github.com/rasbt/mlxtend/files/7463375/1-s2.0-0167865594901279-main.pdf)
 

Here it is compared to the floating version:


Screen Shot 2021-11-02 at 1 52 06 PM
images.githubusercontent.com/5618407/139927438-77579a31-6f2a-48ff-8820-8deecd7eb76e.png)
 
So I would say having SFFS (floating forward) and SFBS (floating backward) is useful. But I can understand if it is too much work in terms of rewriting all the code so having only one of the two is probably sufficient (given that performance is very close anyway)

rasbt:
Finally got around fixing the CI. Then, I was trying to rebase this PR but since it is not in a feature branch but in the master branch itself, I couldn't figure out how. I manually applied all the changes to this PR, and it appears I created quite a mess

rasbt:
Actually not sure why the workflow CI doesn't run. It worked fine in another PR that I just closed a few minutes ago. Maybe it would be easiest to just close this PR, and make a new PR with the 4 files you had initially submitted, i.e,

- columns.py
- generic_selector.py
- strategy.py
- test_generic_selector.py

if you could do this in a separate feature branch instead of master, I think this would make things easier. Let me know if I can help with anything

PR #886, opened 2022-02-02
SequentialFeatureSelection Early Stopping Criterion

aldder:
### Description

According to some studies reported on papers (like this: https://www.researchgate.net/publication/220804144_Overfitting_in_Wrapper-Based_Feature_Subset_Selection_The_Harder_You_Try_the_Worse_it_Gets), the feature selection methodologies known as Wrapper suffer from overfitting as the number of explored states increases.
A method to reduce this overfitting is to use automatic stop criteria (early stop, as the one most known for neural networks).
In this PR I have implemented the criterion of early stopping for the class SequentialFeatureSelector.

2 parameters have been added in during instantiation of the object:
    
    early_stop : bool (default: False)
        Determines whether to prematurely stop execution if the score does not
        improve after a number of iterations set by the `early_stop_rounds`.
        parameter.
    early_stop_rounds : int (default 3)
        Used when `early_stop` is True, it determines the number of iterations
        after which, if no performance boost has been seen, execution is
        stopped.

Code Example:

```
import numpy as np
import pandas as pd
from sklearn.datasets import load_iris
from sklearn.neighbors import KNeighborsClassifier
from mlxtend.feature_selection import SequentialFeatureSelector as SFS
from mlxtend.plotting import plot_sequential_feature_selection as plot_sfs

np.random.seed(0)
iris = load_iris()
X_iris = iris.data
y_iris = iris.target

# add some noies in order to have features to discard
X_iris_with_noise = np.concatenate(
    (X_iris,
    np.random.randn(X_iris.shape[0], X_iris.shape[1])),
    axis=1)

knn = KNeighborsClassifier()
```

```
sfs = SFS(
    estimator=knn,
    k_features=X_iris_with_noise.shape[1],
    forward=True,
    early_stop=False,
    verbose=0)

sfs.fit(X_iris_with_noise, y_iris)
plot_sfs(sfs.get_metric_dict());
```

```
sfs = SFS(
    estimator=knn,
    k_features=X_iris_with_noise.shape[1],
    forward=True,
    early_stop=True,
    early_stop_rounds=2,
    verbose=0)

sfs.fit(X_iris_with_noise, y_iris)
plot_sfs(sfs.get_metric_dict());
```

> ... Performances not improved for 2 rounds. Stopping now!

### Pull Request Checklist

### DIFF
--- a/mlxtend/feature_selection/sequential_feature_selector.py
+++ b/mlxtend/feature_selection/sequential_feature_selector.py
@@ -114,6 +114,14 @@ class SequentialFeatureSelector(_BaseXComposition, MetaEstimatorMixin):
     n_jobs : int (default: 1)
         The number of CPUs to use for evaluating different feature subsets
         in parallel. -1 means 'all CPUs'.
+    early_stop : bool (default: False)
+        Determines whether to prematurely stop execution if the score does not 
+        improve after a number of iterations set by the `early_stop_rounds` 
+        parameter.
+    early_stop_rounds : int (default 3)
+        Used when early_stop is True, it determines the number of iterations 
+        after which, if no performance boost has been seen, execution is 
+        stopped.
     pre_dispatch : int, or string (default: '2*n_jobs')
         Controls the number of jobs that get dispatched
         during parallel execution if `n_jobs > 1` or `n_jobs=-1`.
@@ -178,6 +186,8 @@ class SequentialFeatureSelector(_BaseXComposition, MetaEstimatorMixin):
                  forward=True, floating=False,
                  verbose=0, scoring=None,
                  cv=5, n_jobs=1,
+                 early_stop=False, 
+                 early_stop_rounds=3,
                  pre_dispatch='2*n_jobs',
                  clone_estimator=True,
                  fixed_features=None):
@@ -200,6 +210,14 @@ class SequentialFeatureSelector(_BaseXComposition, MetaEstimatorMixin):
         self.n_jobs = n_jobs
         self.verbose = verbose
         self.clone_estimator = clone_estimator
+
+        if not isinstance(early_stop_rounds, int) or early_stop_rounds < 0:
+            raise ValueError('Number of early stopping round should be '
+                             'an integer value greater than or equal to 0.'
+                             'Got %d' % early_stop_rounds)
+
+        self.early_stop = early_stop
+        self.early_stop_rounds = early_stop_rounds
 
         if fixed_features is not None:
             if isinstance(self.k_features, int) and \
@@ -424,6 +442,8 @@ class SequentialFeatureSelector(_BaseXComposition, MetaEstimatorMixin):
             }
         best_subset = None
         k_score = 0
+        best_score = -np.inf
+        early_stop_count = self.early_stop_rounds
 
         try:
             while k != k_to_select:
@@ -549,6 +569,18 @@ class SequentialFeatureSelector(_BaseXComposition, MetaEstimatorMixin):
                                           custom_feature_names,
                                           X)
                     raise KeyboardInterrupt
+
+                # early stop
+                if self.early_stop and k != k_to_select:
+                    if k_score <= best_score:
+                        early_stop_count -= 1
+                        if early_stop_count == 0:
+                            print('Performances not improved for %d rounds. '
+                                  'Stopping now!' % self.early_stop_rounds)
+                            break
+                    else:
+                        early_stop_count = self.early_stop_rounds
+                        best_score = k_score
 
         except KeyboardInterrupt:
             self.interrupted_ = True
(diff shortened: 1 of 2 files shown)


rasbt:
Thanks for the PR! I agree that overfitting can become an issue. Currently, there is the option

- `k_features="parsimonious"`

which will select the smallest feature set within 1 standard error of the best feature set, which helps with this.

I like adding an `early_stop` option, but I have a few suggestions / concerns regarding the API:

# 1) 

I think that the two parameters 

`early_stop` and  `early_stop_rounds` can be consolidated into a single one. E.g.,

```python

                 if self.early_stop and k != k_to_select:
                     if k_score <= best_score:
```

could be 

```python
                 if self.early_stop_rounds and k != k_to_select:
                     if k_score <= best_score:
```

What I mean is instead of having

- `early_stop : bool (default: False)`
- `early_stop_rounds : int (default 3)`

this could be simplified to 

- `early_stop_rounds : int (default 0)`

# 2) 

The second concern I have is that if a user selects e.g., `k_features=(1, 3), early_stop_rounds=3`, it's not necessarily guaranteed that there will be 1 to 3 features, which can be confusing.  

I wonder if it makes sense to allow `early_stopping_rounds` only for `k_features='best'` and `k_features='parsimonious'`, which both explore the whole feature subset size space. 

E.g., 

- if `k_features='best'` is selected with `early_stop_rounds=0`, it will evaluate and select the best feature subset in the range 1 to m. 
- if `k_features='best'` is selected with `early_stop_rounds=2`, it will evaluate and select the best feature subset in the range 1 to m but it may stop early, which means for forward selection, it may not explore the whole feature subset sizes up to m.

 If a user selects e.g., k_features=3 and `early_stop_rounds=2` we could 

- a) raise an error saying that `f"Early stopping is not supported with fixed feature subset sizes."`
- b) raise a warning saying `f"Early stopping may lead to feature subset sizes that are different from k_features={k_features}."`

What are your thoughts?

aldder:
Thanks for your suggestions, I agree with your points.
I will edit this PR with the followings:
- have only one parameter `early_stop_rounds` : int >= 0
- enable early stopping only if `early_stop_rounds` > 0 and `k_features` in ['best', 'parsimonious']

rasbt:
Sounds great! Looking forward to it!

mlxtend/feature_selection/sequential_feature_selector.py
```diff
@@ -201,6 +207,13 @@ def __init__(self, estimator, k_features=1,
         self.verbose = verbose
         self.clone_estimator = clone_estimator
 
+        if not isinstance(early_stop_rounds, int) or early_stop_rounds < 0:
+            raise ValueError('Number of early stopping round should be '
+                             'an integer value greater than or equal to 0.'
+                             'Got %d' % early_stop_rounds)
```
rasbt, line 213: I think `...Got %d' % early_stop_rounds` might not work if `early_stop_rounds` is not an integer. Maybe it's better to replace it with

`...Got %s' % early_stop_rounds`

mlxtend/feature_selection/sequential_feature_selector.py
```diff
@@ -550,6 +565,20 @@ def fit(self, X, y, custom_feature_names=None, groups=None, **fit_params):
                                           X)
                     raise KeyboardInterrupt
 
+                # early stop
+                if self.early_stop_rounds \
+                        and k != k_to_select \
+                        and self.k_features in {'best', 'parsimonious'}:
```
rasbt, line 571: Instead of the check here, i would maybe change it to raising a `ValueError` in the top of the function if self.k_features is not in {'best', 'parsimonious'} and self.early_stop_rounds. This way the user is aware, and we only have to perform the check once.

rasbt (changes requested):

mlxtend/feature_selection/sequential_feature_selector.py
aldder: Right!

mlxtend/feature_selection/sequential_feature_selector.py
aldder: Do you prefer having this check on top of `fit` function or during initialization?

rasbt: Ahh, totally lost track and missed your comment! Sorry! Regarding your question, I think `fit` might be better to keep it more consistent with scikit-learn behavior.

jimmy927:
What is the status on this ?

I use the k_features="parsimonious" on my model.
But it continues to add more and more features even after it is obvious the model will not improve, and after that it will select one of the very early model anywys.

I think this PR could get my runtime from 10 days into hours ;-)

rasbt:
Thanks for the ping, and I need to look into this some time -- sorry, haven't had a chance recently due to too many other commitments. Unfortunately, I currently don't have a timeline for this.

PR #930, opened 2022-05-05
make isort black compatible

rasbt:


labdmitriy:
Could you please tell is isort/black working ok with this configuration? 
I thought that we need \-\-check (for black) and \-\-check-only (for isort) arguments to return corresponding codes which GitHub Action will use for decision?
I saw that both official GitHub Action for it use these arguments (along with \-\-diff argument).

rasbt:
Thanks for the note! I thought `--check` would be redundant if we use `--diff` but that was not the case. Just tested and updated it.

labdmitriy:
Probably you are tired of my notes, but isort as I understand uses check-only, not check option :)

rasbt:
No worries, I actually find your comments super helpful! It looks like both `check` and `check-only` do the same. My gut feeling is that it used to be `check-only` but they also added `check` for consistency with other tools.

From `isort --help`:

```
  --show-files          See the files isort will be run against with the current config options.
  --df, --diff          Prints a diff of all the changes isort would make to a file, instead of changing it in
                        place
  -c, --check-only, --check
                        Checks the file for unsorted / unformatted imports and prints them to the command line
                        without modifying the file. Returns 0 when nothing would change and returns 1 when the
                        file would be reformatted.
  --ws, --ignore-whitespace
                        Tells isort to ignore whitespace differences when --check-only is being used.
```

labdmitriy:
> No worries, I actually find your comments super helpful! It looks like both `check` and `check-only` do the same. My gut feeling is that it used to be `check-only` but they also added `check` for consistency with other tools.
> 
> From `isort --help`:
> 
> ```
>   --show-files          See the files isort will be run against with the current config options.
>   --df, --diff          Prints a diff of all the changes isort would make to a file, instead of changing it in
>                         place
>   -c, --check-only, --check
>                         Checks the file for unsorted / unformatted imports and prints them to the command line
>                         without modifying the file. Returns 0 when nothing would change and returns 1 when the
>                         file would be reformatted.
>   --ws, --ignore-whitespace
>                         Tells isort to ignore whitespace differences when --check-only is being used.
> ```

Ah sorry, you are right! Thank you.

PR #934, opened 2022-05-09
Fixes #620 Add a warning about non Boolean inputs

### LINKED ISSUE #620: Transition frequent itemset mining algorithms to use Boolean dataframes
As noted in #619, apriori is substantially faster on Boolean types data arrays for certain settings.

It would be good to add a warning about non Boolean inputs and encourage users to use Boolean DataFrames (if we deprecate integer dataframes, we can also simplify the code)

In a similar manner, it'd be good to check how the input dataframe type affects

- `fpgrowth`
- `fpmax`

NimaSarajpoor:
Add warning about not bool type. This PR addresses issue #620.

### DIFF
--- a/mlxtend/frequent_patterns/fpcommon.py
+++ b/mlxtend/frequent_patterns/fpcommon.py
@@ -1,9 +1,12 @@
 import collections
+import warnings
 from distutils.version import LooseVersion as Version
 
 import numpy as np
 import pandas as pd
 from pandas import __version__ as pandas_version
+
+warnings.simplefilter("always", DeprecationWarning)
 
 
 def setup_fptree(df, min_support):
@@ -105,6 +108,11 @@ def valid_input_check(df):
     # Fast path: if all columns are boolean, there is nothing to checks
     all_bools = df.dtypes.apply(pd.api.types.is_bool_dtype).all()
     if not all_bools:
+        warnings.warn(
+            "Support for dataframe with non-bool type migth be discontinued in "
+            "the future. Use dataframe with type bool",
+            DeprecationWarning,
+        )
         # Pandas is much slower than numpy, so use np.where on Numpy arrays
         if hasattr(df, "sparse"):
             if df.size == 0:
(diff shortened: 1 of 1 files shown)


mlxtend/frequent_patterns/fpcommon.py
```diff
@@ -105,6 +108,11 @@ def valid_input_check(df):
     # Fast path: if all columns are boolean, there is nothing to checks
     all_bools = df.dtypes.apply(pd.api.types.is_bool_dtype).all()
     if not all_bools:
+        warnings.warn(
+            "Support for dataframe with non-bool type migth be discontinued in"
```
rasbt, line 112: A little typo: 

- migth -> might

And some little suggestions :)

I would phrase it as 

> Support for DataFrames with non-bool types result in worse computational performance and might be discontinued in the future. Please use a DataFrame with bool type.

rasbt (approved):
Thanks a lot for the PR! Just a small suggestion

mlxtend/frequent_patterns/fpcommon.py
NimaSarajpoor: Thanks for the feedback! I will revise it and push again.

NimaSarajpoor: @rasbt 
I think the term `Support` and `result in worse computational performance` together in one sentence is a little bit weird. What do you think?

I rephrased that as follows but I am not sure if it gets green light from you or not.
```
DataFrames with non-bool types result in worse computational performance and their support might be discontinued in the future. Please use a DataFrame with bool type"
```

rasbt: Ohh, there must have been something gone missing when I copy & pasted the two sections together 😅. I liked your rephrased version actually :)

rasbt:
I just added a Changelog entry. You may have to do a `git pull` before you make any additional commits. Btw. I just checked out the Deprecation warning, and it works for all three algos. Nice! I think after updating the message, it is good to merge!

NimaSarajpoor:
@rasbt 
Thanks for your support and guidance! I thinks things should be good now. 

I did not know that I should have added this small change to `CHANGELOG.md`. So, if I understand correctly, (almost?) any change to this library should be recorded in `CHANGELOG.md` after they are finalized right before merging the PR.

rasbt:
No worries and thanks for the PR. Let me run the tests and then I think it should be good to merge

PR #935, opened 2022-05-11
be more specific about isort usuage

rasbt:


rasbt:
CC @labdmitriy ^^

labdmitriy:
Is it correct that you use biopandas as first party package in isort configuration, not mlxtend?

rasbt:
Ooops. I was working on both packages in parallel ... still crazy that I mixed that up 🙈. Should be fixed via #936

PR #985, opened 2022-11-08
Fixed #971 turned off joblib when `n_jobs == 1`

### LINKED ISSUE #971: Exhaustive Feature Selector Performance is poor
I recently noticed that running the EFS on larger datasets now has a very poor performance due to blocking the CPU. I think this has something to do with the joblib rewrite and sending out the individual tasks.

On one hand, it's to be expected that large feature sets can't be exhaustively samples. On the other hand, EFS now supports KeyboardInterrupt, so it should be in theory possible to just evaluate a unreasonably large number of combinations and then interrupt the run.

A potential way to make this work would be to have an iterative algorithm for feature subset sampling when `n_jobs=0`. Right now, this setting is no supported anyway, so we can take advantage of that.

NimaSarajpoor:
This PR fixes issue #971 

# Performance Code
```
seed = 0
X = np.random.rand(10000, 10) # 10k samples, with 10 features
y = np.random.choice([0, 1], size=10000)

lst = []
for i in range(5):
    tic = time.time()
    efs = EFS(RandomForestClassifier()).fit(X, y) # EFS: ExhaustiveFeatureSelector
    toc = time.time()
    lst.append(toc - tic)

np.mean(lst)
```

# Computing Time
* branch `main`: 103 sec
* this branch: 93 sec

### DIFF
--- a/docs/sources/CHANGELOG.md
+++ b/docs/sources/CHANGELOG.md
@@ -7,6 +7,23 @@ The CHANGELOG for the current development version is available at
 
 ---
 
+
+
+Version 0.22.0dev (TBD)
+
+##### Downloads
+
+- [Source code (zip)](https://github.com/rasbt/mlxtend/archive/v0.22.0.zip)
+
+- [Source code (tar.gz)](https://github.com/rasbt/mlxtend/archive/v0.22.0.tar.gz)
+
+##### Changes
+
+- When `ExhaustiveFeatureSelector` is run with `n_jobs == 1`, joblib is now disabled, which enables more immediate (live) feedback when the `verbose` mode is enabled. [#985](https://github.com/rasbt/mlxtend/pull/985) via [Nima Sarajpoor]
+
+
+
+
 ### Version 0.21.0 (09/17/2022)
 
 
--- a/mlxtend/__init__.py
+++ b/mlxtend/__init__.py
@@ -4,4 +4,4 @@
 #
 # License: BSD 3 clause
 
-__version__ = "0.21.0dev0"
+__version__ = "0.22.0dev0"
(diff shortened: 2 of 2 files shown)


NimaSarajpoor:
I will fix this in the upcoming days.

NimaSarajpoor:
@rasbt 
I think it is ready. if there is something that I missed, please let me know.

rasbt (approved):
thanks a lot, that looks great! Neat & clean solution!

rasbt:
Was just testing the code and it definitely improved the startup time. When I am trying an example like 

```python
import numpy as np
from sklearn.linear_model import LogisticRegression
from mlxtend.feature_selection import ExhaustiveFeatureSelector as EFS

seed = 0
X = np.random.rand(10000, 10) # 10k samples, with 10 features
y = np.random.choice([0, 1], size=10000)

model = LogisticRegression()

efs1 = EFS(model, 
           min_features=1,
           max_features=10,
           scoring='accuracy',
           print_progress=True,
           n_jobs=1,
           cv=5)

efs1 = efs1.fit(X, y)

print('Best accuracy score: %.2f' % efs1.best_score_)
print('Best subset (indices):', efs1.best_idx_)
print('Best subset (corresponding names):', efs1.best_feature_names_)
``` 
it still seems to be a bit stuck though. I.e., it would not show any output for like 2-3 min and then iterate through the 1k possibilities in like 1 sec.

I wonder if that's an issue with the `verbose` display functionality though 🤔

NimaSarajpoor:
> EDIT: No worries, it was a computer issue. It works perfectly now. Actually it solves the problem. Before, a user could not see the progress printed to the command line until all combinations were evaluated. Now, you get the feedback immediately if `n_jobs==1`

Thanks for the info :)

PR #1000, opened 2022-12-01
Fix fpmax for sparse matrices

tkellogg:
### Description

While the current functionality supports sparse matrices on an API level, it calls `pd.DataFrame.values` in order to get the number of rows in the DataFrame. Unfortunately, `.values` forces the entire DataFrame to be converted into a non-sparse 2D numpy array. So using sparse matrices to fix an OOM doesn't actually make the OOM go away. 

This fix is semantically identical, in Pandas terms, except that it doesn't materialize a dense array just to find it's length.

N/A

### Pull Request Checklist

    * NOTE: 5 tests don't pass, but that's consistent with the state prior to this change.


rasbt:
> Unfortunately, .values forces the entire DataFrame to be converted into a non-sparse 2D numpy array

Wow, good catch! Thanks for the PR

Issue #1034, opened 2023-05-14
Leftover use of np.int and np.bool


carlsmedstad:
#### Describe the bug

There is still some places where `np.int` and `np.bool` are used which should be replaces with `np.int_` and `np.bool_` respectively:

```sh
$ grep -R 'np.bool)'
mlxtend/evaluate/holdout.py:        train_mask = np.ones(X.shape[0], dtype=np.bool)
$ grep -R 'np.int)'
docs/sources/user_guide/classifier/OneRClassifier.ipynb:    "    return X_discretized.astype(np.int)\n",
mlxtend/evaluate/bias_variance_decomp.py:            var += (pred != main_predictions).astype(np.int)
```

carlsmedstad:
The following fixes the tests:

```sh
sed -i 's/astype(np.int)/astype(np.int_)/g' mlxtend/evaluate/bias_variance_decomp.py
sed -i 's/dtype=np.bool/dtype=np.bool_/g' mlxtend/evaluate/holdout.py
```

rasbt:
Should be fixed now :)

PR #1040, opened 2023-05-17
Replaced np.int and np.bool

namanmistry:
### Description

As in the issue `Leftover use of np.int and np.bool #1034` I have replaced the `np.int` and `np.bool`  with `np.int_` and `np.bool_` in the `bias_variance_decomp.py` and `holdout.py` files.

Fixes Leftover use of np.int and np.bool #1034

### Pull Request Checklist


rasbt:
Thanks for the fix!

PR #1066, opened 2023-09-21
ci fix

rasbt:


jmahlik:
I wonder if making a fresh new conda environment might be an alternative to updating the base? 

Or pip installing the development requirements in base from a `testing` extra? Like `python -m  pip install .[testing]`? Then you might not have to mess with conda updates at all.

rasbt:
Yeah, I actually intermittently tried pip but then it had some issues with Cython. I wanted to just get it to work again like it used to -- it worked a few months ago. It's really frustrating to not have direct access to the machine and to try it one commit at a time 😅

jmahlik:
Willing to take a look/help out. I do ci stuff a lot at work.

rasbt:
Thanks, I think the problem is that we are using old package versions. And there seems to be some incompatibility with the latest Ubuntu version and and old glib version that is required here. 

In an ideal world, this could be fixed by updating the unit tests to use the latest packages for pandas etc. But this would require other work in the frequent pattern submodules, which I am saving for a rainy day in the future.

It's tricky. Let's see if this older Ubuntu version works first and go from there 😅

rasbt (commented):
bump version

rasbt:
It seemed to be a Cython issue that was fixed 🎉 by bumping the sklearn version @jmahlik

Issue #1004, opened 2022-12-18
mlxtend fpgrowth and association rules with the existence of missing values


zazass8:
#### Describe the workflow you want to enable
I am trying to implement the fpgrowth algorithm on a dataset that I am working on. I do know for a fact that the input fpgrowth only accepts is in binary format (e.g. True, False or 1's,0's). My dataset is in binary format but I have some missing values within it. I can't really drop those missing values because that will result in dropping useful binary datapoints from the rest.

#### Describe your proposed solution
What I thought of doing, is to compute support and confidence metrics based on the frequency of items that exist in that binary map AND by also ignoring the missing values. I believe that would be doable and I did find how it's done from this paper (https://weber.itn.liu.se/~aidvi/courses/06/dm/papers/AssociationRules/RAR(98).pdf). The formulas for support and confidence are quite different, mainly they subtract the count of the missing values from the denominator of the support/confidence. They also include a new metric known as 'representativity' and it is an additional criterium that helps in determining better which association rules are the strongest ones.

#### Describe alternatives you've considered, if relevant
I tried to make a few edits on the open source code you have on github, but I got lost while reading it, it's a very abstract code to be honest. Because, you have been using a lot of recursions as well. 

If it is possible you could help me to make these edits I will very much appreciate it.
#### Additional context

zuari1993:
Hi, 
It would be great if you can explain the exact formula that is mentioned in the paper and the exact changes to be made.
I would be more than happy to take this up and make the changes.

I just don't have the time to read through the paper and identify the needed changes.  Thanks.

zazass8:
Okay, so the old rule of support was given as the proportion of the existence of an item out of all transactions. Now, it is given as
![image](https://user-images.githubusercontent.com/64093265/208707032-d4fd52a2-25f4-4a17-8a84-82368db80184.png) where |B_x| is the count of item X, |B| is the count of transactions of the whole database and |Dis(X)| is the count of null values in the column of item X. 

For confidence, given a rule as X->Y, the old definition was given as the proportion of the count of the existence of X and Y occurring together over the count of the existence of X. Now, the new definition is given as 
![image](https://user-images.githubusercontent.com/64093265/208708595-f672461a-af22-436b-afe0-5b2db91f9ec8.png)
where |B_xy| is the the count of the existence of X and Y occurring together, |B_x| is the count of the existence of X and |Dis(Y) interesection B_x| is the count of null values in the column of item Y AND the value in item X exists.

[...]

PR #1106, opened 2024-10-07
FPGrowth/FPMax and Association Rules with the existence of missing values (#1004)

zazass8:
### Description

This update intends to implement the FP-Growth and FP-Max algorithms from frequent_patterns with the possibility of missing values in the input dataset. The code implements the same structure and logic of the algorithms, while computing the support metric as in "ignoring" the missing values in the data. That gives a more realistic indication of the frequency of existence in the items/itemsets that are generated from the algorithm. Given the output of the algorithm, the corresponding association rules and metrics are also updated taking into account the existence of missing values.

The input accepted for this implementation is a `pandas.DataFrame` that accepts binary input as 0/1 or True/False (as it was originally), as well as 0/1/NaN or True/False/NaN, where NaN is `numpy.nan`.

Please also find attached the corresponding paper conducting this research of association rule mining with the existence of missing values.
[ragel1998.pdf](https://github.com/user-attachments/files/17276897/ragel1998.pdf)

Please use [this link](https://github.com/rasbt/mlxtend/issues/1004) for the corresponding discussion of this new feature in the issue tracker.

### Pull Request Checklist


zazass8:
@rasbt there's an issue with CI, as you can see in fpcommon.py line 34 the function np.nansum() is not passing several tests whereas testing it with np.sum() passes all CI tests but practically it gives the wrong answers. I think it may require to make some changes to the tests. I tried already to change the code in fpommon.py but it seems np.nansum() is the one that is suitable for our purpose. By counting the 1s while "ignoring" NaNs. Let me know if you have in mind an alternative solution.

zazass8:
@rasbt In metrics such as Zhang's and Jaccard, if `null_values=True` it computes values that are outside the unrestricted range. Sometimes it occurs with the certainty metric as well. I haven't found anywhere in the literature if there's a modified formula of these metrics that also takes into account the existence of NaNs. If there is, I am happy to update otherwise if you prefer I could exclude these metrics from the final array that is created. Let me know :)

docs/sources/CHANGELOG.md
```diff
@@ -18,11 +18,18 @@ The CHANGELOG for the current development version is available at
...
+  - [`mlxtend.frequent_patterns.association_rules`](https://rasbt.github.io/mlxtend/user_guide/frequent_patterns/association_rules/)
+- [`mlxtend.frequent_patterns.association_rules`](https://rasbt.github.io/mlxtend/user_guide/frequent_patterns/association_rules/)Implemented three new metrics: Jaccard, Certainty, and Kulczynski. ([#1096](https://github.com/rasbt/mlxtend/issues/1096))
+- Integrated scikit-learn's `set_output` method into `TransactionEncoder` ([#1087](https://github.com/rasbt/mlxtend/issues/1087) via[it176131](https://github.com/it176131))
 
 ##### Changes
 
+- [`mlxtend.frequent_patterns.fpcommon`] Added the null_values parameter in valid_input_check signature to check in case the input also includes null values. Changes the returns statements and function signatures for setup_fptree and generated_itemsets respectively to return the disabled array created and to include it as a parameter. Added code in [`mlxtend.frequent_patterns.fpcommon`] and [`mlxtend.frequent_patterns.association_rules`](https://rasbt.github.io/mlxtend/user_guide/frequent_patterns/association_rules/) to implement the algorithms in case null values exist when null_values is True.
```
rasbt, line 32:
```suggestion
- [`mlxtend.frequent_patterns.fpcommon`] Added the `null_values` parameter in `valid_input_check` signature to check in case the input also includes null values. Changes the returns statements and function signatures for `setup_fptree` and `generated_itemsets` respectively to return the disabled array created and to include it as a parameter. Added code in [`mlxtend.frequent_patterns.fpcommon`] and [`mlxtend.frequent_patterns.association_rules`](https://rasbt.github.io/mlxtend/user_guide/frequent_patterns/association_rules/) to implement the algorithms in case null values exist when `null_values` is `True`.
```

rasbt:
Just wanted to say thanks for the PR! I first bookmarked it 2 weeks ago, but then I got sick and am just slowly catching up.

mlxtend/frequent_patterns/fpgrowth.py
```diff
@@ -42,6 +44,9 @@ def fpgrowth(df, min_support=0.5, use_colnames=False, max_len=None, verbose=0):
       The support is computed as the fraction
       transactions_where_item(s)_occur / total_transactions.
 
+    null_values : bool (default: True)
```
rasbt, line 47:
```suggestion
    null_values : bool (default: False)
```

I think this should say `False` based on the function signature above. (This is a great feature, but `False` is probably also the right choice for backward compatibility.)

mlxtend/frequent_patterns/association_rules.py
```diff
@@ -30,6 +33,9 @@
 
 def association_rules(
     df: pd.DataFrame,
+    df_or: pd.DataFrame,
```
rasbt, line 36: If I see it correctly, the DataFrame is only required if `null_values` is `True`, correct? In this case, we could also make this argument optional and set it to `None` by default for backwards compatibility. (And then if someone saids `null_values=True` we raise an error that `df_or` should be provided.

Perhaps we should also rename `df_or` to `df_orig`. The name `df_or` made be first think of a logical or like `df_and` and `df_or`. What do you think?

zazass8:
No worries at all, my pleasure to contribute on your project! No worries at all, I am glad you are feeling better now. I resolved and made the changes you mentioned above, feel free to have another look and let me know if there's anything else!

mlxtend/frequent_patterns/fpcommon.py
```diff
@@ -163,6 +164,19 @@ def valid_input_check(df, null_values=False):
             "Please use a DataFrame with bool type",
             DeprecationWarning,
         )
+
+        # If null_values is True but no NaNs are found, raise an error
```
rasbt, line 168: Just curious, would this lead to inaccurate metrics or is it just inefficient. Or are there other reasons why this would not be permitted?

mlxtend/frequent_patterns/tests/test_association_rules.py
```diff
@@ -55,7 +55,7 @@
 
 # fmt: off
 def test_default():
-    res_df = association_rules(df_freq_items, df, len(df))
+    res_df = association_rules(df_freq_items, len(df), df)
```
rasbt, line 58: If NaN values are not present, the unit tests don't need the extra `df` here though, correct? I wonder if we could have the unit tests that don't have NaN values as they were before but then have an additional one where `df` is provided as you showed just to check that it doesn't cause errors. What do you think?

rasbt:
> No worries at all, my pleasure to contribute on your project! No worries at all, I am glad you are feeling better now. I resolved and made the changes you mentioned above, feel free to have another look and let me know if there's anything else!

Thanks for addressing these! I just had 2 follow-up comments above, but no rush.

mlxtend/frequent_patterns/fpcommon.py
zazass8: No, this one is just inefficient. Normally if I don't have an array with null values but null_values is True, it'd generate the final dataframe with metrics when null values do not exist (as original approach). The only reason I included is to remain consistent and still raise the error if it is the case. But since null_values is set False as default, maybe there's no point of including this condition as the user may not intend to set null_values to True if there are no NaNs in their array. Up to you, if you want I can completely exclude it.

mlxtend/frequent_patterns/tests/test_association_rules.py
zazass8: We can definitely do that, it is a great idea! I removed `df` from the signature but still kept `num_itemsets=len(df)` as it is necessary even for the test case when NaNs do not exist. I also added a `test_nullability()` method to test the case when null values exist. Let me know if you need anything else.

mlxtend/frequent_patterns/fpcommon.py
rasbt: Thanks, that's what I thought! Sounds good! In this case, maybe we should not raise an error but a warning that this may be inefficient, and that `null_values` can be set to False for a speed-up. What do you think?

zazass8: Fixed! Please look at the latest commit.

mlxtend/frequent_patterns/tests/test_association_rules.py
rasbt: Awesome, thanks for adding this. I'd say don't worry about the formatting. It's one of the caveats when using black. It's just in the tests so it's ok.

mlxtend/frequent_patterns/fpcommon.py
```diff
@@ -104,14 +148,36 @@ def valid_input_check(df):
...
             "DataFrames with non-bool types result in worse computational"
             "performance and their support might be discontinued in the future."
             "Please use a DataFrame with bool type",
             DeprecationWarning,
         )
+
+        # If null_values is True but no NaNs are found, raise an error
+        has_nans = pd.isna(df).any().any()
+        if null_values and not has_nans:
+            warnings.warn(
```
rasbt, line 171: Looks great, thanks for the update!

rasbt (approved):
Awesome work! And thanks so much for addressing all the minor concerns I have.

zazass8:
> Awesome work! And thanks so much for addressing all the minor concerns I have.

No worries at all, my pleasure in contributing. Until next time!

d-kleine:
fixed #1004

jraymartinez:
Memory usage regression after PR #1106

We noticed significant memory usage increase in our production pipeline using fpgrowth algorithm after recent updates. 

Details:
- Pipeline was running successfully weekly until November 8, 2024
- Same input data size and parameters throughout (no changes in data volume or fpgrowth parameters)
- Originally running on ml.c5n.9xlarge (96GB RAM)
- After failure, upgraded to ml.c5n.18xlarge (192GB RAM) but still getting memory errors
- Error message: "Please use an instance type with more memory, or reduce the size of job data processed on an instance"
- When pinned to mlxtend==0.23.1 (pre-PR version), pipeline runs successfully again on original instance size

This suggests potential memory regression after PR #1106 (merged Oct 23, 2024). Could you please investigate if there are any unintended memory usage increases from these changes?

Environment:
- AWS SageMaker
- Python 3.10
- Input data shape: (1832691, 11161)
- FPGrowth parameters: min_support=0.0005, max_len=15

rasbt:
Thanks for letting us know. I will take a look at this in the next few weeks and see what's causing this. In the meantime, I recommend sticking to `mlxtend==0.23.1` as you suggested.

If you have any ideas @zazass8 let me know

zazass8:
I investigated the issue. It seems that in `fpcommon.py` where the `disabled` array is initialised, multiple temporary arrays are created to generate the desired one. For that I already worked on an alternative solution which it could work.

Furthermore, on line 40 where I used `np.sum` and `np.logical_or` would consume a lot of memory for the same as above. For this, when I was working on the PR I actually used directly `np.nansum()` as I did in the next line. Where, theoretically it can solve the memory allocation issue as it directly computes what is desired. Although, several unit tests when the code was tested on CI were failing without me figuring out the source of the issue. 

Finally, in `generate_itemsets` maybe the fact that multiple if statements are implemented may cause again memory allocation. I couldn't think of any other way to simplify those lines in order to obtain the desired output.

@rasbt if you have any ideas, let me know as well. Thank you for pointing this out @jraymartinez.

rasbt:
Thanks for the analysis @zazass8! I would say we should maybe investigate why `np.nansum` doesn't/didn't work. Maybe that is the main culprit here.
I currently only have access to a small laptop until December but can help try some things in December when I am back.

billyes:
Why do we add num_itemsets as required input for association_rules? It can be inferred from df size so at least we should make it optional IMO.

zazass8:
> Why do we add num_itemsets as required input for association_rules? It can be inferred from df size so at least we should make it optional IMO.
The reason it was added is because if we use the algorithm when the input dataset includes missing values the metrics that are generated along with the association rules are updated differently (please have a look at the updated formulae on `metric_dict`). You are right, it shouldn't be required for the case if `null_values=False` but I would suggest to instead set it as default equal to 1 in order to remain consistent with the implementation of the formulae. Then if you do this, also remove line 288 (`num_itemsets = 1`) because I noticed it is not really needed if we set it as default.

PR #1133, opened 2025-12-09
Add tol option

mariam851:
### Add 'tol' option for termination criterion in SequentialFeatureSelector

Hi Sebastian,

I added a new `tol` option to SequentialFeatureSelector so that users can stop the feature selection early 
if the improvement is smaller than the tolerance. 

This works similarly to sklearn's tol parameter.

All tests run successfully locally. Happy to adjust anything if needed!

Thank you for the great library and examples!


rasbt (changes requested):
Thanks for the PR. Overall, this looks good to me. However, I'd say these comments are unnecessary as it's clear what's changed in the file diff. There is no need to explicitly comment any changes like that.

rasbt:
Thanks for all the updates, I appreciate it. I totally forgot and missed to check on this. Thanks again for doing this.

rasbt (approved):

PR #1155, opened 2026-01-10
Fix #982: update CI and environment dependencies with proper formatting

### LINKED ISSUE #982: tests are not passing in the branch `main`
It seems that the test functions for `feature_selection` are not passing in the branch `main`.

I got:
```
# run in bash: pytest mlxtend/feature_selection/tests/.

####

================= 24 failed, 71 passed, 15 warnings in 34.95s =================

```

mariam851:
Hi @rabst, 
### Description
This PR addresses the issue #982 where test failures occurred in the `feature_selection` module due to a dependency conflict between **NumPy 2.4.0** and **Numba**.

### Changes
- Updated `environment.yml` to pin NumPy to `>=1.21.0,<2.4.0`.
- Modified `.github/workflows/python-package-pip.yml` and `python-package-conda.yml` to ensure stable NumPy versions during CI runs.
- Verified that all 95 tests in `mlxtend/feature_selection/` pass successfully in a local environment.


rasbt:
Thanks for the PR! Maybe I missed (or forgot) something, but why is numba added?

mariam851:
Hi @rasbt, you're right. numba isn't a direct dependency of mlxtend. I added it to the Conda workflow because of the compatibility issues between NumPy 2.x and older Numba versions which were causing failures during environment setup in the CI.

However, since NumPy is now pinned to <2.4.0, we probably don't need to install numba explicitly anymore. I can remove it from the workflow to keep the dependencies minimal. Would you like me to do that?

.github/workflows/python-package-conda.yml
```diff
@@ -29,7 +29,8 @@ jobs:
       - name: Install test dependencies (conda)
         shell: bash -l {0}
         run: |
-          conda install -y pytest coverage
+          # Pinning numpy and scikit-learn to prevent the issues reported in #982
+          conda install -y pytest coverage "numpy>=1.22,<2.4" "scikit-learn>=1.0" "numba>=0.59.0"
```
rasbt:
```suggestion
          conda install -y pytest coverage "numpy>=1.22,<2.4" "scikit-learn>=1.0"
```

.github/workflows/python-package-pip.yml
```diff
@@ -24,6 +24,8 @@ jobs:
       - name: Install dependencies (pip)
         run: |
           python -m pip install --upgrade pip
+          # Fixing #982 by pinning numpy to a version compatible with numba
```
rasbt:
```suggestion
          # Fixing #982 by pinning numpy
```

rasbt (approved):
> Hi @rasbt, you're right. numba isn't a direct dependency of mlxtend. I added it to the Conda workflow because of the compatibility issues between NumPy 2.x and older Numba versions which were causing failures during environment setup in the CI.

I see, thanks for explaining, that makes sense. I removed it here in the PR to see if it works, and if not, we can add it.

rasbt:
Actually, I am just looking at #982 and it's from 2022. This one has been addressed a long time ago. I don't think we need to change anything because there was no failure in main.

PR #1168, opened 2026-05-21
fix(preprocessing): map constant columns to min_val in minmax_scaling (#1167)

### LINKED ISSUE #1167: minmax_scaling returns NaN silently for constant columns
#### Describe the bug

#### Steps/Code to Reproduce

```python
import numpy as np

ary = np.array([[5, 1],
                [5, 2],
                [5, 3]])

print(result)
```

#### Expected Results

```
[[0.  0. ]
 [0.  0.5]
 [0.  1. ]]
```

#### Actual Results

```
[[nan  0. ]
 [nan  0.5]
 [nan  1. ]]
```

#### Versions

```
MLxtend 0.23.4
Windows-11-10.0.26200-SP0
Scikit-learn 1.7.0
NumPy 2.2.1
SciPy 1.15.2
```

SAY-5:
### Code of Conduct

I have read the project's Code of Conduct.

### Description

`minmax_scaling` silently returns `NaN` for constant columns because the `(max - min)` denominator is zero. This change substitutes the zero denominator with 1.0, so the numerator (also zero on those columns) divides cleanly and constant columns end up at `min_val`, matching how `standardize` already handles its zero-variance case.

Fixes #1167

### Pull Request Checklist

- [ ] Modify documentation in the corresponding Jupyter Notebook under `mlxtend/docs/sources/` (not applicable, behavior matches expectation in the issue)
- [x] Ran `PYTHONPATH='.' pytest ./mlxtend/preprocessing -sv` and all tests pass (40/40 including 2 new regression tests)
- [x] Checked for style issues by running `flake8 ./mlxtend` and `ruff check` (clean)

### DIFF
--- a/docs/sources/CHANGELOG.md
+++ b/docs/sources/CHANGELOG.md
@@ -25,6 +25,8 @@ The CHANGELOG for the current development version is available at
 
 - `bias_variance_decomp` now accepts pandas DataFrames and Series as input, in addition to NumPy arrays. ([#1070](https://github.com/rasbt/mlxtend/issues/1070) via [berns722](https://github.com/berns722))
 
+- Fix `preprocessing.minmax_scaling` returning `NaN` for constant columns; constant columns are now mapped to `min_val` instead of triggering a `0 / 0` divide. ([#1167](https://github.com/rasbt/mlxtend/issues/1167))
+
 
 ### Version 0.24.0  (13 Dec 2025)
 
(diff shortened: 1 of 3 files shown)


rasbt:
Thanks for the PR. Looks like there were lots of duplicates. Since [#1169](https://github.com/rasbt/mlxtend/pull/1169) is already merged, this one would not be needed anymore. Thanks though!

PR #1170, opened 2026-05-21
evaluate: clarify avg_bias is squared bias in bias_variance_decomp (#1083)

### LINKED ISSUE #1083: “Bias squared” rather than “Bias”?
I have a question about the “Average bias” in [this documentation](https://rasbt.github.io/mlxtend/user_guide/evaluate/bias_variance_decomp/) and [the coding of the bias_variance_decomp function](https://github.com/rasbt/mlxtend/blob/master/mlxtend/evaluate/bias_variance_decomp.py).

If Loss = [Bias]^2 + Variance, then in the reporting of the bias_variance_decomp function, it should be **average bias squared** rather than **average bias**?

Any comment is welcome, many thanks!

jbbqqf:
## Summary

The `bias_variance_decomp` docstring describes the second return value as "average bias", but for the `mse` loss the function actually computes the average **squared** bias — `mean((main_predictions - y_test) ** 2)` — which is the `Bias^2` term in the standard `Loss = Bias^2 + Variance` decomposition. This mismatch is exactly the point raised in #1083, which the maintainer agreed should be clarified in the docs.

This PR is docstring-only (plus one inline code comment); no runtime behaviour changes. The variable name `avg_bias` is kept for backwards compatibility.

Fixes rasbt/mlxtend#1083 — *"Bias squared" rather than "Bias"?*

## Context

Two small problems in the current `Returns` section of `bias_variance_decomp`:

1. The text reads `"the average expected average bias, and average bias (all floats)"` — a copy-paste typo where `average variance` is duplicated as `average bias`.
2. There is no hint that, for the `mse` loss, `avg_bias` is the squared bias rather than the raw bias. Readers who match it against `Loss = Bias^2 + Variance` (as the original reporter did) expect a square root or an explicit note.

The maintainer confirmed in [#1083](https://github.com/rasbt/mlxtend/issues/1083#issuecomment-1825036569): *"Yes, that's a good point. It's average biased squared. Maybe it should be made more explicit in the docs like you suggest."*

## Changes

- `mlxtend/evaluate/bias_variance_decomp.py` — rewrote the `Returns` block to spell out the three values, fix the duplicated-variance typo, and add an explicit note about `avg_bias` being squared bias for `mse` (and the misclassification rate of the main prediction for `0-1_loss`). Also added a 4-line code comment next to the `mse` `avg_bias` computation so a reviewer reading the diff cold doesn't have to re-derive the invariant.
- `docs/sources/CHANGELOG.md` — one entry under "Version 0.25.0 (TBD)".

## Reproduce BEFORE/AFTER yourself (copy-paste)

This change is documentation-only, so the BEFORE/AFTER is the docstring rendering:

```bash
# --- one-time setup ---
git clone https://github.com/rasbt/mlxtend.git /tmp/repro-1083 && cd /tmp/repro-1083
python -m venv .venv && source .venv/bin/activate
pip install -e .

# --- BEFORE (origin/master) ---
git checkout origin/master
python -c "from mlxtend.evaluate import bias_variance_decomp; help(bias_variance_decomp)" | sed -n '/Returns/,/Examples/p'
# Expected: the Returns section reads
#   "avg_expected_loss, avg_bias, avg_var : returns the average expected
#    average bias, and average bias (all floats), where the average is
#    computed over the data points in the test set."
# i.e. "average bias" is duplicated and there is no Bias^2 vs Bias note.

# --- AFTER (this PR) ---
git fetch https://github.com/jbbqqf/mlxtend.git fix/1083-bias-squared-docs
git checkout FETCH_HEAD
python -c "from mlxtend.evaluate import bias_variance_decomp; help(bias_variance_decomp)" | sed -n '/Returns/,/Examples/p'
# Expected: the Returns section now lists "average expected loss, average
# bias, and average variance" and explicitly notes that for 'mse' the
# avg_bias is the squared bias, the term that appears as Bias^2 in
# Loss = Bias^2 + Variance.
```

## What I ran locally

```text
$ pytest mlxtend/evaluate/tests/test_bias_variance_decomp.py -v --timeout=60
========================= 5 passed, 1 failed in 6.06s ==========================
```

The single failure is `test_keras` skipping locally because TensorFlow isn't installed (`ModuleNotFoundError: No module named 'tensorflow'`); the test is GitHub-Actions-skipped via `@pytest.mark.skipif(TRAVIS or APPVEYOR or GITHUB_ACTIONS, …)`, so it'll be skipped on CI too. The five non-Keras tests all pass.

## Edge cases tested

| # | Scenario | Input | Expected | Verified by |
|---|----------|-------|----------|-------------|
| 1 | Numeric behaviour unchanged | existing tests for `0-1_loss` and `mse` | identical floats vs main | `test_bias_variance_decomp.py` (5/5 pass) |
| 2 | Docstring renders | `help(bias_variance_decomp)` after install | new `Returns` block visible | manual check shown above |

## Risk / blast radius

Documentation-only on the public function plus an inline code comment. No callers' behaviour changes; the return tuple shape and the values inside it are byte-for-byte identical.

## Release note

```release-note
Clarified in the `bias_variance_decomp` docstring that `avg_bias` for the `mse` loss is the average *squared* bias (the Bias^2 term in `Loss = Bias^2 + Variance`), and fixed a typo where the `Returns` section read "average bias, and average bias" (#1083).
```

---

*PR drafted with assistance from Claude Code. The clarification matches the maintainer's own comment in #1083; the variable-name preservation was decided to avoid breaking downstream tuple-unpacking callers. The reproducer block above was used during development; it is the same one a reviewer can paste verbatim.*

### DIFF
--- a/docs/sources/CHANGELOG.md
+++ b/docs/sources/CHANGELOG.md
@@ -25,6 +25,8 @@ The CHANGELOG for the current development version is available at
 
 - `bias_variance_decomp` now accepts pandas DataFrames and Series as input, in addition to NumPy arrays. ([#1070](https://github.com/rasbt/mlxtend/issues/1070) via [berns722](https://github.com/berns722))
 
+- Clarified in the `bias_variance_decomp` docstring that, for the `mse` loss, `avg_bias` is the average *squared* bias (the `Bias^2` term in `Loss = Bias^2 + Variance`), and fixed a typo in the `Returns` section that previously read "average bias, and average bias". Behaviour unchanged. ([#1083](https://github.com/rasbt/mlxtend/issues/1083) via [jbbqqf](https://github.com/jbbqqf))
+
 
 ### Version 0.24.0  (13 Dec 2025)
 
--- a/mlxtend/evaluate/bias_variance_decomp.py
+++ b/mlxtend/evaluate/bias_variance_decomp.py
@@ -68,8 +68,17 @@ def bias_variance_decomp(
     Returns
     ----------
     avg_expected_loss, avg_bias, avg_var : returns the average expected
-        average bias, and average bias (all floats), where the average
-        is computed over the data points in the test set.
+        loss, average bias, and average variance (all floats), where the
+        average is computed over the data points in the test set.
+
+        Note that for the ``'mse'`` loss, ``avg_bias`` reports the
+        average **squared** bias
+        (``mean((main_predictions - y_test) ** 2)``) — i.e. the term
+        that already appears squared in the standard bias-variance
+        decomposition ``Loss = Bias^2 + Variance``. For the
+        ``'0-1_loss'``, ``avg_bias`` is the misclassification rate of
+        the main prediction (the majority vote across bootstrap
+        replicates), which is the 0-1 analogue of squared bias.
 
     Examples
     -----------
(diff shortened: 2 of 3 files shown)


rasbt:
Thanks for improving this, I appreciate it!

PR #1171, opened 2026-05-24
fix: raise ValueError in minmax_scaling for constant columns

### LINKED ISSUE #1167: minmax_scaling returns NaN silently for constant columns

karthikankam:
## Summary
Fixes the silent NaN bug in `minmax_scaling()` when a constant 
column (zero range) is passed.

## Problem
When all values in a column are identical:
- denominator = max - min = 0
- result = 0/0 = NaN (silent, only a RuntimeWarning is emitted)
- This can corrupt ML pipelines without any clear indication

## Fix
Added a zero-range check before the division:

```python
are_constant = denominator == 0
if np.any(are_constant):
    raise ValueError(
        "One or more columns have zero range (all values are identical). "
        "minmax_scaling cannot be applied to constant columns."
    )
```

## Why ValueError over silent 0.0
- Silent behavior hides data issues from the user
- Explicit errors are safer and easier to debug in ML pipelines
- Consistent with how standard libraries handle invalid input

## Testing
```python
import numpy as np

# Constant column → raises ValueError
ary = np.array([[5,1],[5,2],[5,3]])
minmax_scaling(ary, columns=[0,1])  # ValueError ✅

# Normal case → still works
ary2 = np.array([[1,1],[2,2],[3,3]])
minmax_scaling(ary2, columns=[0,1])  # [[0. 0.][0.5 0.5][1. 1.]] ✅
```

Closes #1167


rasbt:
Looks good, thanks! I just added a test case for this.

PR #1185, opened 2026-06-30
validate proportion inputs in proportion_difference

Qayad-Ali:
proportion_difference silenlty returned (nan, nan) (with a RuntimeWarning) when a proportion was outside [0, 1], for  e.g. proportion_difference(1.5, 0.3, 50). This adds validation that raises a clear ValueError when proportion_1 or proportion_2 is not in [0, 1] — consistent with the input checks already in mcnemar and cochrans_q. Valid-input behavior is unchanged; a unit test was added.


rasbt (approved):
LGTM. Thanks for the PR

PR #1191, opened 2026-08-28
apriori: make low_memory match the default result at exact support boundaries

AlejandroCoronadoN:
`apriori(..., low_memory=True)` can silently drop an itemset whose support lands exactly on `min_support`, while the default path keeps it. The two paths are documented as producing the same result (`low_memory` is described only as a slower, memory-efficient mode).

The default path compares the normalized support: `count / rows_count >= min_support`. The `low_memory` path instead scaled the threshold (`threshold = min_support * rows_count`) and compared raw counts against it. That multiplication can round up: for 25 rows and `min_support=0.28`, `0.28 * 25 == 7.000000000000001`, so an itemset with a count of exactly 7 (support `7/25 = 0.28`) fails `7 >= 7.000000000000001` and is dropped.

Repro:

```python
import pandas as pd
from mlxtend.frequent_patterns import apriori
from mlxtend.preprocessing import TransactionEncoder

tx = [['A', 'B']] * 7 + [['C']] * 18   # {A, B} support = 7/25 = 0.28
te = TransactionEncoder()
df = pd.DataFrame(te.fit(tx).transform(tx), columns=te.columns_)

apriori(df, min_support=0.28)                    # includes {A, B}
apriori(df, min_support=0.28, low_memory=True)   # drops {A, B}
```

The fix compares the normalized support directly (`supports / rows_count >= min_support`), matching the default path's arithmetic so the two modes agree. The yielded support values are unchanged. Added a regression test at the exact boundary; the existing `test_low_memory_matches_default_result` uses data that does not land on a float boundary, so it did not catch this.


rasbt (approved):
Awesome, great catch and great PR! Many thanks!

PR #1193, opened 2026-09-02
Correct documented default values that disagree with the signatures

VenishPaneliya:
### Code of Conduct

Read and agreed.

### Description

Eight parameter defaults are documented with a value the signature does not use:

| function | parameter | documented | actual |
|---|---|---|---|
| `lift_score` | `positive_label` | `0` | `1` |
| `paired_ttest_kfold_cv` | `shuffle` | `True` | `False` |
| `find_filegroups` | `validity_check` | `None` | `True` |
| `fpmax` | `null_values` | `True` | `False` |
| `plot_confusion_matrix` | `figsize` | `(2.5, 2.5)` | `None` |
| `plot_confusion_matrix` | `norm_colormap` | `False` | `None` |
| `heatmap` | `figsize` | `(2.5, 2.5)` | `None` |
| `checkerboard_plot` | `figsize` | `(2.5, 2.5)` | `None` |

The first four affect results rather than appearance:

- `positive_label` selects which class the lift is computed for, so a reader who trusts the docstring expects the score for class `0`.
- `shuffle` changes the CV split the paired t-test runs on.
- `validity_check` gates a check that raises (`if validity_check:` in `find_filegroups`).
- `null_values` is threaded through `valid_input_check`, `setup_fptree` and the generator.

For `figsize`, neither `plot_confusion_matrix` nor `heatmap` actually falls back to `(2.5, 2.5)` — both scale with the matrix:

```python
if figsize is None:
    figsize = (len(conf_mat) * 1.25, len(conf_mat) * 1.25)
```

and `checkerboard_plot` passes `None` straight through to matplotlib, so the effective size is matplotlib's own default rather than `(2.5, 2.5)`. `norm_colormap` reaches `ax.matshow(norm=...)`, which expects a `Normalize` instance or `None`, not a bool.

Only docstrings are changed — no behaviour is touched. I checked the rest of the package for the same problem; the other near-matches (`find_filegroups.substring`, `scatterplotmatrix.figsize`, `scatter_hist.figsize`, `standardize.params`, `plot_decision_regions.feature_index`) are all correct as written and are left alone.

None.

### Pull Request Checklist

- [ ] Added appropriate unit test functions in the `./mlxtend/*/tests` directories (if applicable) — not applicable, docstring-only
- [ ] Modify documentation in the corresponding Jupyter Notebook under `mlxtend/docs/sources/` (if applicable) — not applicable, the notebooks render these docstrings via the API docs
- [x] Ran the relevant test files and made sure the unit tests pass (`mlxtend/evaluate/tests/test_lift_score.py`, `mlxtend/frequent_patterns/tests/test_fpmax.py`: 26 passed)
- [x] Checked for style issues (`black`, `isort` and `flake8` all clean on the changed files)


rasbt (approved):
Great, thanks for the PR fixing these doc issues.

PR #1195, opened 2026-09-07
Fix plotting docstrings that name parameters the functions do not have

VenishPaneliya:
Two plotting docstrings name parameters that don't exist, so following them raises `TypeError`. Docs only, no behaviour change.

### `heatmap` documents `conf_mat`, parameter is `matrix`

```python
def heatmap(
    matrix,
    hide_spines=False,
    ...
```

```
    Parameters
    -----------
    conf_mat : array-like, shape = [n_rows, n_columns]
        And arbitrary 2D array.
```

`conf_mat` is the parameter name used by `plot_confusion_matrix`, so this looks like the docstring was adapted from that function and the name wasn't updated — the description right below it ("And arbitrary 2D array") reads like it was rewritten for the general case while the name stayed behind. Every other entry in the docstring matches the signature. I also corrected "And arbitrary" to "An arbitrary" on that same line.

### `scatterplotmatrix` documents `fix_axes`, parameter is `fig_axes`

```python
def scatterplotmatrix(
    X, fig_axes=None, names=None, figsize=(8, 8), alpha=1.0, **kwargs
):
```

The docstring says `fix_axes` in both the Parameters and the Returns sections, but it's inconsistent with itself — the `figsize` entry a few lines further down already refers to `fig_axes` correctly:

```
    figsize : tuple (default: (8, 8))
      Height and width of the subplot grid. Ignored if
      fig_axes is not `None`.
```

so `fix_axes` is a typo for `fig_axes` rather than a rename. Fixed both occurrences.

Both functions are public and have user-guide pages, so these are what the rendered API docs show.

Checked with `black --check` — both files unchanged.


rasbt (approved):
Oh good catch, thanks for the PR!

